PR #8, opened 2022-09-06
Generate hugo input

reid-spencer:
Make a little scalatest suite to generate hugo output with `sbt test`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,11 @@
+Global / onChangedBuildSource := ReloadOnSourceChanges
 scalaVersion := "2.13.8"
+// enablePlugins(RiddlSbtPlugin)
 libraryDependencies ++= Seq(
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -0,0 +1,11 @@
+Global / onChangedBuildSource := ReloadOnSourceChanges
+scalaVersion := "2.13.8"
+// enablePlugins(RiddlSbtPlugin)
+libraryDependencies ++= Seq(
+  "com.reactific" %% "riddl-testkit" % "0.13.+" % "test",
```
JamesYopp, line 5: Biggest take away here is I need to remember to update this when we update RIDDLC...

JamesYopp (approved):
Looks like there are a couple of new(?) places that I need to remember to update when the riddl version updates (build.sbt and project/plugins.sbt).

JamesYopp (approved):

PR #14, opened 2022-09-14
Upgrade workflow to 0.14.1

reid-spencer:
* Everything else stays at local build 0.14.+
* Simplified build.sbt syntax to only change riddl version in 1 place


JamesYopp (dismissed):

reid-spencer:
I simplified it a lot, and it validated the model in 13 seconds! Merging!

PR #17, opened 2022-09-17
Fixes for new state syntax (riddlc 14.2)

JamesYopp:
Projections need a fix from RIDDLC for this to work. Otherwise I think we are now compatible with 14.2.


reid-spencer (commented):
LGTM

PR #20, opened 2022-10-05
IA-73 Venues/Location/Reservations

AlexWeinstein92:


src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
```diff
@@ -0,0 +1,23 @@
+type Status is any of {Draft, OnHold, Reserved, Active, Past} briefly "Internal status of a Reservation."
+
+type Info is {
```
JamesYopp, line 3: Does Reservation need to have a location: LocationId attribute? A reservation is for an Organization, during TimeRange, at Location...

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
```diff
@@ -0,0 +1,23 @@
...
+    range: TimeRange,
+    title: String,
+    description: String,
+    forOrg: OrganizationId,
+    isPrivate: Boolean briefly "Is this reservation accessible outside of the organization owning it?"
```
JamesYopp: I think we mark the event private. I am not sure what effect marking a reservation private would do.

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
JamesYopp, line 3: I wonder if reservation also need to have a reference to an Event? A reservation is made by an organization, during TimeRange, at Location, for Event

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
```diff
@@ -0,0 +1,23 @@
...
+}
+
+type UpdateInfo is {
```
JamesYopp, line 11: should the UpdateInfo type have all the attributes of Info?

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
```diff
@@ -0,0 +1,23 @@
...
+    forOrg: OrganizationId?
+}
+
+type MetaInfo is {
+  createdOn: TimeStamp briefly "The Date/Time this venue was created.",
+  createdBy: MemberId briefly "The member who created this venue",
+  lastUpdated: TimeStamp briefly "The Date/Time this venue was last modified.",
+  lastUpdatedBy: MemberId briefly "The member who last modified this venue. This may also be a system user. i.e. they are two months late on payment so the system determines to suspend the venue.",
+  currentStatus: InternalStatus briefly "The current state of this venue. Valid states include Draft, Active, Suspended, and Terminated",
+  currentOwner: MemberId
```
JamesYopp, line 26: I wonder if the "owner" is the organization associated with the reservation?

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
```diff
@@ -0,0 +1,23 @@
...
+    range: TimeRange?,
+    forOrg: OrganizationId?
+}
+
+type MetaInfo is {
+  createdOn: TimeStamp briefly "The Date/Time this venue was created.",
+  createdBy: MemberId briefly "The member who created this venue",
+  lastUpdated: TimeStamp briefly "The Date/Time this venue was last modified.",
+  lastUpdatedBy: MemberId briefly "The member who last modified this venue. This may also be a system user. i.e. they are two months late on payment so the system determines to suspend the venue.",
+  currentStatus: InternalStatus briefly "The current state of this venue. Valid states include Draft, Active, Suspended, and Terminated",
```
JamesYopp: States/Status are Draft, OnHold, Reserved, Active, Past, right?

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
+command EstablishReservation is {
```
JamesYopp, line 1: Maybe MakeReservation? I think that is how our users would call this action...

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
...
+  reservationId: ReservationId briefly "A unique identifier for this Reservation. This must be provided by the calling client.",
+  info: ReservationContext.Info briefly "Details about a Reservation required for it to exist. See ReservationContext.Info for details.",
+  establishingMember: MemberId briefly "The Id of the Member who is issuing the create command.",
```
JamesYopp: ReservingMember?

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
...
+  owner: MemberId? briefly "Used in the case that establishingMember is not the owner"
```
JamesYopp: I wonder if Owner should be the organization making the reservation?

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
...
+} briefly "Creates a Reservation"
+
+event ReservationEstablished is {
```
JamesYopp, line 8: ReservationMade?

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
...
+command AddReservationAttendees is {
+  reservationId: ReservationId,
+  attendees: MemberId+
+}
+event ReservationAttendeesAdded is {
+  venueId: VenueId,
+  allAttendees: MemberId+
+}
+
+command DeleteReservation is {reservationId: ReservationId}
```
JamesYopp, line 47: CancelReservation/ReservationCancelled?

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
...
+  reservationId: ReservationId,
+  status: ReservationContext.Status,
+  updatingMember: MemberId
+}
+event ReservationStatusUpdated is {
+  reservationId: ReservationId,
+  newStatus: ReservationContext.Status
+}
+
+command AddReservationAttendees is {
```
JamesYopp, line 35: I don't see an attendees attribute? Where would one add them to?
I think Attendees is a complication that we should avoid right now. 
(future)I think maybe instead of attendees on an reservation we should have invitations to an event.

src/main/riddl/Venues/Reservations/reservationMessages.riddl
```diff
@@ -0,0 +1,45 @@
...
+  owner: MemberId? briefly "Used in the case that establishingMember is not the owner"
+} briefly "Creates a Reservation"
+
+event ReservationEstablished is {
+  reservationId: ReservationId,
+  info: ReservationContext.Info,
+  meta: ReservationContext.MetaInfo
+}
+
+command EditReservationInfo is {
```
JamesYopp, line 14: I think this could be ChangeReservation or UpdateReservation
I think we could handle updating all details of a Reservation in a single call like this. For Organization we updated specific fields where there was high frequency (addMembers/removeMembers) or high complexity (UpdateParent, AddContacts, etc.), so I thought it necessary to have an UpdateInfo command that changed a block of Info rather than having a bunch of getters/setters for things that didn't fall in the above categories.

src/main/riddl/Venues/Reservations/reservation.riddl
```diff
@@ -0,0 +1,115 @@
...
+		            reservationMeta: ReservationContext.MetaInfo
+		        }
+		        handler ActiveReservation is {
+		            on command EstablishReservation {
+		                then error "Reservation already established"
+		            }
+		            on command EditReservationInfo {
+		                then error "Reservation is in Active state, editing info not allowed"
+		            }
+		            on command UpdateReservationStatus {
```
JamesYopp, line 94: I think the only state we could transition to from Active is Past, correct? We might want to list examples. See organization.riddl for examples. Briefly here:
```
state Active is {
  handler ActiveReservation is {
    on command UpdateReservationStatus {
      example NewStatusIsDraft is {
        given "Reservation current status is Active"
        when "newStatus is Draft"
        then "error: Active reservations cannot transition back to a Draft state"
      }
   }
} 
```

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
```diff
@@ -0,0 +1,24 @@
...
+	name: String,
+  description: String,
+  directions: String briefly "How do you find this location?",
+  latLong: LatLong,
+  url: URL?,
+  capacity: Integer?,
+  accessible: Boolean briefly "Is this location handicapped accessible?",
+  amenities: String*, //TODO: We may need amenities to be searchable - i.e. I need to find a Location that is handicaped accessible and has a kitchen
+  entrances: Address+,
+  isPrivate: Boolean briefly "Is this location accessible outside of the organization owning its venue?"
```
JamesYopp, line 15: To be accessible (or not) outside of the owning organization we would need to have reference to an Organization ID

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type MetaInfo is {
+  createdOn: TimeStamp briefly "The Date/Time this venue was created.",
+  createdBy: MemberId briefly "The member who created this venue",
+  lastUpdated: TimeStamp briefly "The Date/Time this venue was last modified.",
+  lastUpdatedBy: MemberId briefly "The member who last modified this venue. This may also be a system user. i.e. they are two months late on payment so the system determines to suspend the venue.",
+  currentStatus: LocationContext.Status briefly "The current state of this venue. Valid states include Draft, Active, Suspended, and Terminated",
+  currentOwner: MemberId
```
JamesYopp, line 37: Should current owner be an Organization?

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
```diff
@@ -0,0 +1,24 @@
...
+	id: LocationId,
+	name: String,
+  description: String,
+  directions: String briefly "How do you find this location?",
+  latLong: LatLong,
+  url: URL?,
+  capacity: Integer?,
+  accessible: Boolean briefly "Is this location handicapped accessible?",
+  amenities: String*, //TODO: We may need amenities to be searchable - i.e. I need to find a Location that is handicaped accessible and has a kitchen
+  entrances: Address+,
```
JamesYopp, line 13: If there are multiple entrances do we need a preferred entrance?

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
```diff
@@ -0,0 +1,24 @@
...
+type Info is {
+	id: LocationId,
+	name: String,
+  description: String,
+  directions: String briefly "How do you find this location?",
+  latLong: LatLong,
+  url: URL?,
+  capacity: Integer?,
+  accessible: Boolean briefly "Is this location handicapped accessible?",
+  amenities: String*, //TODO: We may need amenities to be searchable - i.e. I need to find a Location that is handicaped accessible and has a kitchen
```
JamesYopp, line 12: Should amenities be an enumerated list of strings? Searching becomes harder if somebody misspells stuff (accessalbe, Kithcen, etc.)

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
```diff
@@ -0,0 +1,24 @@
...
+
+type Info is {
+	id: LocationId,
+	name: String,
+  description: String,
+  directions: String briefly "How do you find this location?",
+  latLong: LatLong,
+  url: URL?,
+  capacity: Integer?,
+  accessible: Boolean briefly "Is this location handicapped accessible?",
```
JamesYopp: Is accessible an amenity? Or is it its own attribute? Your comment in the next line indicates that it is amenity.

src/main/riddl/Venues/Locations/locationMessages.riddl
```diff
@@ -0,0 +1,58 @@
...
+  name: String,
+  description: String,
+  directions: String,
+  latLong: LatLong,
+  url: URL?,
+  capacity: Integer?,
+  accessible: Boolean,
+  amenities: String*,
+  entrances: Address+,
+  reservations: ReservationId*
```
JamesYopp, line 12: I think you should model reservations for a Location as a projection. Reservations by Location is important to query, but is not defining for a Location.

src/main/riddl/Venues/Locations/locationMessages.riddl
```diff
@@ -0,0 +1,58 @@
+command EstablishLocation is {
+  locationId: LocationId briefly "A unique identifier for this Location. This must be provided by the calling client.",
+  name: String,
+  description: String,
+  directions: String,
+  latLong: LatLong,
+  url: URL?,
+  capacity: Integer?,
+  accessible: Boolean,
+  amenities: String*,
```
JamesYopp, line 10: several attributes defined here are defined again in the Info type. I think we should remove the redundancy.

src/main/riddl/Venues/Locations/locationMessages.riddl
```diff
@@ -0,0 +1,58 @@
...
+} briefly "Creates a Location with all top-level fields, potentially without Locations"
+
+event LocationEstablished is {
+  locationId: LocationId,
+  locationInfo: LocationContext.Info,
+  reservations: ReservationId*,
+  meta: LocationContext.MetaInfo
+}
+
+command AddLocationReservations is {
```
JamesYopp, line 22: I don't think we should have AddLocationReservation. I think this should be handled in the Reservation Context and you get Reservations for a Location by query against a Projection.
As a side note, just as a general principle, if you have an AddReservation you should also provide a RemoveReservation. If we nix AddLocationReservation as suggested, then of course, you would not have a remove operation...

src/main/riddl/Venues/Locations/locationMessages.riddl
```diff
@@ -0,0 +1,58 @@
...
+  locationId: LocationId,
+  reservations: ReservationsId+
+}
+event LocationReservationsAdded is {
+  locationId: LocationId,
+  allReservations: ReservationId+
+}
+
+query GetNextLocationReservations is {
```
JamesYopp, line 52: I think these Reservation/Availability queries are good. Again, we need to specify the projection(s) that would best support these queries.

src/main/riddl/Venues/venueTypeDefinitions.riddl
```diff
@@ -0,0 +1,37 @@
+type Status is any of {Draft, Open, Closed} briefly "Internal status of a Venue."
+
+type Info is {
+	orgId: OrganizationId briefly "Venues are defined by BaseOrgs. venueId and orgId make a composite key",
+  name: String,
+  address: Address,
+  isPrivate: Boolean briefly "Is this location accessible outside of the organization owning its venue?"
```
JamesYopp: You would need an organizationId to see if it falls outside of the organization. I have made several of these organization contexts, owners being organizations etc. But a thought just struck me this time (good thing it finally did). This idea of organization ownership is good. I think it is interesting to have a private location/venue too. But to enforce this private idea you would need to recurse down the organization hierarchy to see if a reserving org belongs to the owning org (a generational child). Let me know if this idea does not make sense.

JamesYopp: I see OrgId in the info block above, but will leave the comment to capture the idea of the generational child and private Venues.

src/main/riddl/Venues/venueTypeDefinitions.riddl
```diff
@@ -0,0 +1,37 @@
...
+}
+
+type UpdateInfo is {
+	orgId: OrganizationId?,
+  name: String?,
+  address: Address?
+}
+
+type VenueReservationsSchedule is {
```
JamesYopp, line 17: I don't think the idea of a reservation schedule belongs here. I think there is a projection that would provide this info.

src/main/riddl/Venues/venueTypeDefinitions.riddl
```diff
@@ -0,0 +1,37 @@
...
+  maxCapacity: Integer
+}
+
+type MetaInfo is {
+  createdOn: TimeStamp briefly "The Date/Time this venue was created.",
+  createdBy: MemberId briefly "The member who created this venue",
+  lastUpdated: TimeStamp briefly "The Date/Time this venue was last modified.",
+  lastUpdatedBy: MemberId briefly "The member who last modified this venue. This may also be a system user. i.e. they are two months late on payment so the system determines to suspend the venue.",
+  currentStatus: VenueContext.Status,
+  currentOwner: MemberId
```
JamesYopp, line 38: should currentOwner be of type OrganizationId?

src/main/riddl/Venues/venueMessages.riddl
```diff
@@ -0,0 +1,79 @@
...
+	venueId: VenueId,
+	maxCount: Integer,
+	startDate: DateTime? briefly "Used as start date of query when present instead of now()"
+}
+result VenueAvailabilityCalendar is {
+	venueId: VenueId,
+	schedule: VenueAvailabilitySchedule* briefly "A list of location with time ranges indicating availability"
+}
+
+query GetVenueCapacityOverview is {
```
JamesYopp, line 85: I can't see where CapacityOverview is used in an Entity/State definition. Either we need to add that attribute to the fields of an Entity/State or we remove the definition and the associated queries.

JamesYopp: I did not read the briefly... Doh!
I like the idea of a projection here. Where is the projection definition?

src/main/riddl/Venues/venueMessages.riddl
```diff
@@ -0,0 +1,79 @@
...
+command AddVenueLocations is {
+  venueId: VenueId,
+  locations: LocationId+
+}
+event VenueLocationsAdded is {
+  venueId: VenueId,
+  allLocation: LocationId+
+}
+
+query GetNextVenueReservations is {
```
JamesYopp, line 45: I think VenueReservations need to be provided by a projection. I think the queries are fine.

src/main/riddl/Venues/venueMessages.riddl
```diff
@@ -0,0 +1,79 @@
...
+  venueId: VenueId,
+  status: VenueContext.Status,
+  updatingMember: MemberId
+}
+event VenueStatusUpdated is {
+  venueId: VenueId,
+  newStatus: VenueContext.Status
+}
+
+command AddVenueLocations is {
```
JamesYopp, line 36: I think VenueLocations should be maintained as a projection. The Location needs to have an attribute of VenueId to declare which venue it belongs to. I think the natural thought process would be, "I am going to add a new Location". When setting up that Location you would need to identify the Location it belongs to, if any. The LocationAdded/LocationRemoved events would update the projection.

src/main/riddl/Venues/venueContext.riddl
```diff
@@ -0,0 +1,106 @@
...
+                and "yield event VenueEstablished"
+            }
+            on other { then error "Only EstablishVenue command is allowed in the default state"}
+        }
+
+		    state Draft is {
+		        fields {
+		            venueId: VenueId,
+		            venueInfo: VenueInfo,
+		            locations: LocationId*,
```
JamesYopp, line 23: I don't think we should define Locations in Venue

src/main/riddl/Venues/venueContext.riddl
```diff
@@ -0,0 +1,106 @@
...
+                    and "yield event VenueInfoUpdated"
+		            }
+			      }
+	      }
+
+		    state Open is {
+		        fields {
+		            venueId: VenueId,
+		            venueInfo: VenueInfo,
+		            locations: LocationId+ briefly "Venues must have at least 1 location to be considered Available",
```
JamesYopp: I don't think we should have locations in Venue

src/main/riddl/Venues/venueContext.riddl
```diff
@@ -0,0 +1,106 @@
...
+                    and "yield event VenueInfoUpdated"
+		            }
+		        }
+		    }
+
+		    state Closed is {
+		        fields {
+		            venueId: VenueId,
+		            orgId: OrganizationId,
+		            locations: LocationId*,
```
JamesYopp, line 79: I don't think we should have Locations in Venue

src/main/riddl/Venues/venueMessages.riddl
```diff
@@ -0,0 +1,79 @@
+command EstablishVenue is {
+  venueId: VenueId briefly "A unique identifier for this Venue. This must be provided by the calling client.",
+  info: VenueContext.Info briefly "Details about a venue required for it to exist. See VenueContext.Info for details.",
+  locations: LocationId*,
```
JamesYopp: I don't think we should have locations in venue

src/main/riddl/Venues/venueMessages.riddl
```diff
@@ -0,0 +1,79 @@
...
+  establishingMember: MemberId briefly "The Id of the Member who is issuing the create command.",
+  owner: MemberId? briefly "Used in the case that establishingMember is not the owner"
```
JamesYopp, line 5: Is the owner an OrganizationId?

src/main/riddl/Venues/venueMessages.riddl
```diff
@@ -0,0 +1,79 @@
...
+} briefly "Creates a Venue with all top-level fields, potentially without Locations"
+described in file "establishVenue.md"
+event VenueEstablished is {
+  venueId: VenueId,
+  info: VenueContext.Info,
+  locations: LocationId*,
```
JamesYopp: I don't think Venues should have Locations

src/main/riddl/Venues/venueTypeDefinitions.riddl
```diff
@@ -0,0 +1,37 @@
...
+type VenueReservationsSchedule is {
+	locationId: LocationId,
+	reservations: ReservationId+
+}
+type VenueAvailabilitySchedule is {
+	locationId: LocationId,
+	schedule: TimeRange+
+}
+
+type CapacityOverview is {
```
JamesYopp, line 27: If CapacityOverview is provided as a projection do we need this type?

reid-spencer:
@AlexWeinstein92 - I merged my changes to `main` branch into this branch. They are unrelated but they do fix the validation workflow to use the right version of riddlc. So, once you've resolved the comments from @JamesYopp it should run the checks again. Currently it would use a version that does not have the sorted output turned off and may likely fail like it was when we were testing earlier today.  I will work tonight to make that not be the case so that if errors do show up, at least the main culprit is the first one printed.

src/main/riddl/Venues/venueMessages.riddl
reid-spencer: Why? A location is something inside a Venue, right?  So, how did you want to express the relationship?

JamesYopp: Long story. But part of the complexity I have found while modeling this app is relationships. Venues has a collection of Locations. A Location belongs to a Venue. I think one "side" or the other of the relationship needs to own the data about the relationship. If both own the data, because of isolation, you run into a multi-master situation which is difficult to keep synchronized. It also makes logic in the handlers more complex. So pick one side of the relationship. I picked the Location side to own the Venue<->Location relationship. The other side of the relationship can be obtained by using a projection. This projection listens for LocationAdded events and updates the projection if the Location specifies a Venue.

Why did I pick the location side to "own" the relationship? In this case, it is the simple side of a 1:M relationship. I don't need to parse a list to figure out how to update the other side of the relationship. 

Another factor I am considering is an idea I am calling a "defining attribute". You need to maintain data in some way on both sides of a relationship - either as a field or within a projection. If the linking attribute is optional I would say that data is not defining. So that entity is a poor candidate for "owning" the relationship. In this specific case, a Venue must have at least one Location. A Location, on the other hand, does not need to specify a Venue. Using this logic Venue would be the better candidate to own the relationship because at least Location is required. 

I weighted the cardinality of the 1:M relationship and decided that modeling Location as the "owner" of the relationship made the overall model simpler. I am very open to more conversation on this as I think it is an important topic that will help future authors of RIDDL code considerably.

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
AlexWeinstein92: This brings up the question of 'what is the difference between Reservation and Event?'. I'm not exactly sure what you had in mind. When I was showing him my code yesterday, Reid was surprised that Reservation was even being included in this context. Is it mostly a financial administration thing or is there more to it?

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
AlexWeinstein92: I was thinking Owner has to belong to associated Organization, and represents highest level admin. We could make it a list if that makes more sense.

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
AlexWeinstein92: Yes

src/main/riddl/Venues/Reservations/reservationMessages.riddl
AlexWeinstein92: If we are going that realistic I think BookReservation would be appropriate

src/main/riddl/Venues/Reservations/reservationMessages.riddl
AlexWeinstein92: Outdated - not needed because we can just use UpdateStatus

src/main/riddl/Venues/Reservations/reservationMessages.riddl
AlexWeinstein92: I put Attendees on the State definitions instead of in Info because it is likely to be changing often. 

Wondering where you see the complications arising? Invites indeed would would be more complicated, but unless we want to have lots of restrictions & admin tooling around controlling who can be an Attendee, it seems as simple as giving Members ability to register to Events. Maybe this is related to the difference between Events & Reservations?

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
AlexWeinstein92: My idea was to go to Venue that owns this Location to make this field effective. Maybe what we need is a reference to the Venue this is linked to?

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
JamesYopp: A Reservation is simply a contract to be able to use a particular place at a particular time. As long as it is not illegal and it fits with the terms of use established by the Venue/Location, I am not sure that the "owner" cares what happens. I think eventually there could be a financial aspect to a Reservation. You pay for use. If the reservation is a contract to use, the Event is what is happening during that time. A play, a game, a class, a working session, etc. People get invited to an event not a reservation. Does that help clear things up?

src/main/riddl/Venues/Reservations/reservationMessages.riddl
JamesYopp: Sorry, I missed it in the state definitions. My mistake. I think the observation of complications still stands. I think eventually it makes sense, but I think it is future scope. I don't think it adds to the effectiveness of this first version we are rolling out (the comparison study). To be a full featured product I think it will need to be there. Again, future scope, I think there is a related concept to an Invitation, and that is a Ticket. You sell tickets to an event. When you buy one you are "invited". In a way, a Ticket is kind of like a Reservation. I am securing a seat at an event. A reservation I am securing a Location to hold an event.

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
JamesYopp: You would need something. Let's think about frequency of use (every time somebody makes a reservation or looks for availability) and what would need to happen to successfully enforce this business rule. Also, I would ask if this feature is important to the Comparison Study? Do we need it now or should it be future scope?

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
AlexWeinstein92: Amenities will probably need to be their own entities. This is left over from the original VenueContext that had Location and Reservation in the same file as Venue. I can get rid of it until we see Amenities being worth building out

src/main/riddl/Venues/venueTypeDefinitions.riddl
AlexWeinstein92: It's a good thing to point out. We would need to write a query on isPrivate connecting it to a projection based on parents & children of OrgId. I don't think that would be too difficult to build out though. Will mention it in the comment.

src/main/riddl/Venues/venueMessages.riddl
AlexWeinstein92: I haven't got around to projections yet... but I will :) I partly wanted you to look this over as a go-ahead to create projections for these queries, wasn't sure if they could be better designed

src/main/riddl/Venues/venueMessages.riddl
AlexWeinstein92: "LocationAdded/LocationRemoved" as in the creation events?

JamesYopp: Yeah, this comment was all kinds of confusing. Sorry. Location needs to have a VenueId (if there is one) to set up the relationship between Location and Venue. So when a Location is created we would update the project. Thinking about this for a minute more, the projection would only want to show Locations that were in the "Open" state. Draft and Closed Locations probably wouldn't be so helpful. Thoughts?

src/main/riddl/Venues/venueMessages.riddl
AlexWeinstein92: If Location is 'defining attribute' of Venue-Location relationship, Venue's State relies on the projection of that relationship - as long as we are consistent with a requirement I came up with that a Venue cannot be outside Draft State without at least one Location.

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
AlexWeinstein92: Makes sense

AlexWeinstein92: So should we be including legal aspects like Terms of Use in our Venue, Location, and Reservation definitions right now? How will that shape out - because I feel like it is better, but more difficult maybe, to provide a way to restrict who books and what they can do through the app, rather than than build out some some of automatic notification mechanism for indicators of illegality

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
AlexWeinstein92: It could be useful for Comparison Study because complex projections are good test cases, and I think the complexity of writing this query is potentially lower than the complexity of how it will run, That being said, it could be left out from a purely Domain perspective

src/main/riddl/Venues/venueMessages.riddl
AlexWeinstein92: Sure, I think that works. I mean we can always adjust the projection to include Draft and Closed if needed by users (if we ever want/expect any). One idea that comes to mind is caching the projection only for Open state. That may speed things up, without making the other data unavailable.

src/main/riddl/Venues/venueMessages.riddl
JamesYopp: What I was trying to say is I think one side of the relationship needs to be authoritative for the data defining the relationship. I was trying to explain my logic for how I arrived at Location being that thing for the Venue<->Location relationship. I considered many things (including "defining attributes") but settled on the side of simplicity; the 1 side of the 1:M relationship to own the relationship data. I think the state transition rules would have to consult the projection to ensure we are in compliance.

I am still working these design ideas out in my mind. Keep challenging me. It helps sharpen those thoughts.

src/main/riddl/Venues/Reservations/reservationTypeDefinitions.riddl
JamesYopp: No. Future scope for sure.

src/main/riddl/Venues/venueMessages.riddl
JamesYopp: I was suggesting what you observed in the last sentence. The projection only contains (caches) the Locations that are in Open state. So the projection is updated as Locations transition in and out of the Open state.

JamesYopp: I don't think you want to keep ALL Locations in the projection with their current state. That seems to cross some important lines that I can't articulate at the moment (spidey sense)

src/main/riddl/Venues/Reservations/reservation.riddl
AlexWeinstein92: An interesting thing I want to point out: if we are doing this, we need to have a gateway that is always conscious of who initiated a cascading command. Say this command is issued as an effect of another command, and we have to return the error to the user. Or better (or worse) yet, take the situation with Location being master over the VenueLocations projection. If a user makes a new Location with an invalid VenueId, we have to listen for these errors coming back and return them to user for clarity.

This also begs the question - how are we sending automatic updates? Updating projections automatically by listening for messages is a PubSub/Kafka thing, not a GRPC thing.

JamesYopp: Messaging would be done using PubSub/Kafka/NATS/Pulsar/.....(some other similar thing). And you are correct on the error thing. With Async messaging between components the caller would have no other choice but to listen for success/failure. Architecture is all about tradeoffs. You want scale, performance, resilience, reliability, etc.? You need to figure out how to rethink your UX and how to handle communications. If you are not so worried about those things synchronous communication may be the right choice.

AlexWeinstein92: We could create a projection in the Gateway service that facilitates all of the tracing and returning. It would have to be updated on every message send and received by gateway, basically event sourcing (which we can possible leverage from Akka?). It would be pretty complex, but we can build it so it isn't a bottleneck. The real bottleneck is that we would have to rely on Firebase notifications or something similar to get to the actual device that issued the command.

AlexWeinstein92: Actually, we would also have to undo commands that led to these events too. We don't want Location with bad VenueId in the Location service, would be better to wipe it before sending the error back to client, since the operation was ultimately not a success.

JamesYopp: I have a different understanding of event sourcing. As I understand it, event sourcing is the practice of maintaining a log of the events that happen in a system. This log is the authoritative source of the state of an entity. In a "traditional" system the source is the current state. For example, If I have an entity Account with an attribute balance, and that balance is $68.23 that is all I know.

In an event sourced system I have a log of events:
id.      Action.      Amount.     Balance
1234 Deposited $100.00     $100.00
1235 Deposited $33.87       $133.87
1236 Withdrew  $53.27       $  80.60
1237 Deposited $25.00       $105.60
1238 Withdrew  $37.37        $ 68.23

I think either the front end or the gateway (more likely) would need to subscribe to the event stream to listen for events and reply to the device/client.

AlexWeinstein92: Let's say in your example that one of the `Withdraw` events was initiated as a mistake - the wrong account was specified, and it withdrew from this account instead of some other account. You would have to issue another action to deposit that money back into the account by changing order payment information before shipping goes out. The system recognizes that the payment information changed and initiates 2 actions - `Withdraw` to new account and `Deposit` to old account.

The same would apply to creating a new Location with a VenueId that does not exist. Furthermore, Gateway will have no way to associate the `ActionID` of the `UpdateVenueWithLocation` action with that of the `CreateLocation` action. Granted, it can search through the event logs for the `CreateLocation` that contains the  `LocationId` specified in `UpdateVenueWithLocation` . But I don't there's there's any assurance we will always be have enough of an identifying factor.

AlexWeinstein92: We may not be using the best example. When `UpdateVenueWithLocation` fails we can have Gateway ask App for a different `VenueId` and send another `UpdateVenueWithLocation` as well as an `UpdateLocationWithVenue`, or else they can delete their new `Location`.

It does come into play though for the automatic availability query I'm writing. They are very much up in the air right now (see my comments here https://github.com/improving-app/riddl/pull/23), but it is necessarily a multistep operation where we ask for availability from a projection that joins `Location`s to their `Reservation`s, and then go to `ReservationContext` to update the state of the randomly-chosen-from-search-results `ReservationId`... The success message returned by `ReservationContext` will need to use its `LocationId` and `ReservationId` and `timestamp` to find the originating `User` in the logs...

So never mind about this :) interesting stuff to think about, but not really an issue

PR #23, opened 2022-10-13
Jtl gateway (WIP)

AlexWeinstein92:


AlexWeinstein92:
I don't think it's appropriate for the Availability queries to be in either LocationContext or ReservationContext - rather, they will have to depend on a join of both. Which Context, then, should the queries be found in?

AlexWeinstein92:
My Reservation stories are actually duplicating a lot of the Availability stories, both in the Example and Interaction portions, particularly when it involves picking from an Availability list. It would be nice to be able to use one Interaction or Example in another test case, but maybe we are just supposed to assume some setup has been done already (like checking for Availability)? That would greatly simplify the Reservation stories, to just picking an ID from a list that is already in App

AlexWeinstein92:
Also, given that booking a reservation requires that the reservation exists in non-booked state, I am not sure what the use of these Location Availability queries are, other than for simulating browsing loads. In those queries we decided to simply return a list based on either a time range or a maxCapacity, but in the Reservation Availability queries we are assuming that a Reservation has already been created, but not booked, and is being used in the query. For this reason I want to add a story to Reservations that can transition a Reservation into Booked state, assuming it has been found a Location already. But then I am not sure, in the other stories where a list is returned or automatically selected from, whether we should be updating our Reservation with the chosen Location before updating it to Booked status. @reid-spencer this is largely based on your requirement that we use the exact same object type for PreBooked Reservations and Booked Reservations.

PR #27, opened 2022-12-13
Kalix study Tenants

AlexWeinstein92:


src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+        state InProgress is {
+            fields{
+                eventId: EventId,
+                eventInfo: EventInfo,
+                eventMetaInfo: EventMetaInfo
+            }
+            handler InProgressEventHandler is {
+                on command ScheduleEvent{ ??? // is this allowed or does an event have to be delayed first?
```
JamesYopp: This command should throw an error when in the InProgress state.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+                }
+                on command ChangeEventInfo{ ??? // can info be changed while event is in progress? Would this be more likely to confuse or help users?
```
JamesYopp: This command should throw an error when in the InProgress state.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+                }
+                on command CancelEvent{
```
JamesYopp, line 83: This command should throw an error when in the InProgress state.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+									then "morph Event entity to EventStatus.Cancelled"
+                  and "set entity lastModifiedBy to Event.cancellingMember"
+                  and "set entity lastModifiedOn to today()"
+                }
+                on command RescheduleEvent{ ??? // is this allowed or does the event have to be delayed first?
```
JamesYopp: I think it has to be delayed before you can reschedule.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+                }
+                on command DelayEvent{ ??? // is this allowed or return error?
```
JamesYopp: Yes. You can delay an InProgress event. I think InProgress may be the only state that you can delay an event in. In other states you would just reschedule.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+        state Delayed is {
+            fields {
+                eventId: EventId,
+                eventInfo: EventInfo,
+                reason: String,
+                eventMetaInfo: EventMetaInfo
+                //finalScore
+            }
+            handler DelayedEventHandler is {
+                on command ScheduleEvent{
```
JamesYopp, line 116: I think we should throw an error if they try to schedule a Delayed event. We should direct them to reschedule if that is what they are trying to do.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+									then "morph Event entity to EventStatus.InProgress"
+                  and "set entity lastModifiedBy to Event.delayingMember"
+                  and "set entity lastModifiedOn to today()"
+                }
+                on command ChangeEventInfo{
+                  then "set entity eventInfo to message info"
+                  and "set entity lastModifiedBy to Event.cancellingMember"
```
JamesYopp: We may want to look at the attributes for this command. Is it cancellingMember?

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+                  and "set entity lastModifiedOn to today()"
+                }
+                on command RescheduleEvent{
+                  then "set entity evenInfo.expectedStart to Event.start"
+                  and "set entity evenInfo.expectedEnd to Event.end"
+                  and "set entity lastModifiedBy to Event.reschedulingMember"
+                  and "set entity lastModifiedOn to today()"
+                }
+                on command DelayEvent{
```
JamesYopp, line 138: Because we are already in a delayed state I think this should throw an error.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+                  and "set entity lastModifiedOn to today()"
+                  // what about Event.reason? Do we want to store it somewhere?
+                }
+                on command StartEvent{
+                  then "morph Event entity to EventStatus.InProgress"
+                  and "set entity lastModifiedBy to Event.startingMember"
+                  and "set entity lastModifiedOn to today()"
+                }
+                on command EndEvent{
+									then error "End has not started. Cannot end an event that has not started."
```
JamesYopp: minor typo. "Only events that are InProgress may be ended. If you wish to end this event it must be started first. If you wish to cancel the event, please use the CancelEvent Command."

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+        state Past is {
+            fields {
+                eventId: EventId,
+                eventInfo: EventInfo,
+                eventMetaInfo: EventMetaInfo
+                //finalScore
+            }
+            handler PastEventHandler is {
```
JamesYopp, line 160: I think there is the option to say something like:
`on command any {error "Event is completed. Commands are not allowed on this event."}`
I would have to look up the exact syntax.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+        }
+        state Cancelled is {
+            fields {
+                eventId: EventId,
+                eventInfo: EventInfo,
+                eventMetaInfo: EventMetaInfo
+                //reason
+            }
+            handler CancelledEventHandler is {
+                on command ScheduleEvent{ ??? // is this allowed or return error?
```
JamesYopp: Error. Point the user to the RescheduleEvent command.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -0,0 +1,197 @@
...
+                }
+                on command ChangeEventInfo{then error "Event is cancelled. Cannot change info for a cancelled event."}
+                on command CancelEvent{then error "Event is cancelled. Cannot cancel a cancelled event."}
+                on command RescheduleEvent{ ???
```
JamesYopp: Yes, user should be able to reschedule a cancelled event.

src/main/riddl/Gateway/gatewayStories.riddl
```diff
@@ -48,6 +48,20 @@ As an Owner of an Organization I would like to find a Location, Reserve the Loca
...
 
+story EventBookingEasyButton is {
+  case primary is {
+    step from actor Owner "presses easy button to make a reservation" to
+      input ImprovingApp.Improving_app.MyEventsPages.easy
+      briefly "make a reservation by searching for a location - includes max attendance, start/end time, already created event info"
+    step from output ImprovingApp.Improving_app.MyEventsPage.show
+      "presented" to actor Owner
+      briefly "event will now appear as scheduled in a venue that honours max attendance and event info parameters in the nearest time to start and end available"
+  }
```
JamesYopp, line 65: I think there should be a step in here where the user selects a Location from the list returned from the search?

src/main/riddl/Gateway/gatewayStories.riddl
```diff
@@ -48,6 +48,20 @@ As an Owner of an Organization I would like to find a Location, Reserve the Loca
...
+} briefly "A story about scheduling an event with the easy button in Improving.app"
+described as "This is the story of how an event is scheduled with an easy button."
+
+//Alex's Q - should there be more interaction with the app here? Do we want to display a selection of options?
```
JamesYopp: I think we should show a selection of potential locations based on the search parameters. The user then selects the Location they would like to complete the reservation.

src/main/riddl/Events/Reservations/reservationTypeDefinitions.riddl
```diff
@@ -4,15 +4,14 @@ type Info is {
     range: 'TimeRange',
     title: String,
     description: String,
-    location: LocationId,
```
JamesYopp, line 7: Why wouldn't we allow somebody to update the location of an event?

src/main/riddl/KalixStudy.riddl
```diff
@@ -5,13 +5,13 @@
 domain ImprovingApp by author James is {
     include "authors"
     include "Common/Contact"
-    include "Common/UniqueIds"
```
JamesYopp: Why did we remove UniqueIds?

src/main/riddl/LocationsEventsProjections/locationsEventsProjections.riddl
```diff
@@ -0,0 +1,71 @@
...
+
+  projection LocationsEventsView is {
+     fields {
+        orgId: OrganizationId,
+        locationId: LocationId,
+        capacity: Integer,
+        reservationId: ReservationId,
+        reservationStart: TimeStamp,
+        reservationEnd: TimeStamp,
+        isPrivate: Boolean // we can use this for a sum of location, reservation, & venue privacy
```
JamesYopp: I don't entirely understand the comment here. But, it makes me nervous. When you say a sum of location, reservation, and venue privacy, are you suggesting anding the isPrivate flags from all three of these entities in the view? If that is the case it makes me wonder what the purpose of this projection is. What queries are we anticipating to execute against this projection?

To me the reason to have a view that joins locations and events is to derive a list of events grouped by location. Since we are looking at list of events I would primarily be interested in which events in that list are public and which are private. The isPrivate flag for reservation and venue would be irrelevant. The isPrivate flag for a venue is used limit which organizations can schedule the venue. I need to review reservation. I am not sure whether or not reservation needs an isPrivate flag. A private event limits which members have visibility and who can attend.

src/main/riddl/LocationsEventsProjections/locationsEventsProjections.riddl
```diff
@@ -0,0 +1,71 @@
...
+        on event LocationReservationsAdded {
+            then "find or create rows with reservationId set to LocationReservationsAdded.reservationId"
+            and "set rows' locationId to LocationReservationsAdded.locationId"
+        }
+        on event LocationInfoUpdated {
+            then "find rows where locationId is LocationInfoUpdated.locationId"
+            and "set rows' capacity to LocationInfoUpdated.info.capacity"
+            and "set rows' isPrivate to true if LocationInfoUpdated.info.isPrivate is true"
+        }
+        on query ScheduleEasyEvent {
```
JamesYopp: A query should not be able to create a reservation or set attributes on a reservation. In the on event blocks is where we would update data on the projection. The results of the query would inform the caller what actions to take - most likely commands being issued to specific entities - which would emit events that would update this projection.

src/main/riddl/Tenant/tenant.md
```diff
@@ -1,2 +1,4 @@
 The concept of Tenant in Improving.App is simply the finanically responsible entity for all things that 
-are created under that Tenant - Organizations, Members, Events, Venues, Locations, etc.
\ No newline at end of file
+are created under that Tenant - Organizations, Members, Events, Venues, Locations, etc. It is also the
+primary source of truth for all data in the app - every Organization, Member, Event, etc. must belong to
```
JamesYopp: I don't think that tenant is the primary source of all data in the app. The Organization entity is the primary source of data for the Organization. Member for Member, and Event for Event. Organization must reference a valid Tenant, and Event must reference a valid Organization. But Member need not belong to a specific Tenant, Organization, or Event.

src/main/riddl/Tenant/tenant.riddl
```diff
@@ -134,22 +36,31 @@ context TenantContext is {
...
-                on command ActivateTenant {
+                on command TenantContext.EditInfo {
+                    then "Copy fields from TenantContext.EditInfo into Draft info fields"
+                    and "adjust meta fields to reflect the change set"
+                    and "yield event TenantContext.InfoUpdated"
+                }
+                on command TenantContext.UpdateStatus {
                     example Success {
-                        then "Morph Tenant to Active state"
+                        then "Morph Tenant to desired state"
```
JamesYopp: The valid states for a Tenant are Draft, Active, and Suspended. We are in the draft handler. The state machine for a tenant moves from Draft -> Active <-> Suspended. Perhaps we need to define our messages as such. Or maybe in the DraftHandler if we try to move to suspended we throw an error? In other words, the only valid state transition from Draft is to Active.

src/main/riddl/Tenant/tenant.riddl
```diff
@@ -134,22 +36,31 @@ context TenantContext is {
...
-                        and "yield TenantActivated event"
+                        and "yield TenantContext.StatusUpdated event"
+                    }
+                    example SameStatus {
+                        when "Tenant already is in desired state"
+                        then error "Tenant is already in desired state"
                     }
                     example NoBillingAccount {
-                        when "BillingAccount is null"
+                        when "Desired status is Active and BillingAccount is null"
```
JamesYopp: As I read this the thought that pops to mind is it seems you are trying to make a generic UpdateStatus command, when in reality we are only dealing with the Draft state. I think we can be much narrower in our handler definitions given this observation.

src/main/riddl/Venues/Locations/locationTypeDefinitions.riddl
```diff
@@ -12,11 +12,13 @@ type Info is {
...
   preferredEntrance: Address,
-  isPrivate: Boolean briefly "Is this location accessible outside of the organization owning its venue?"
+  isPrivate: Boolean briefly "Is this location accessible outside of the organization owning its venue?",
+  reservations: ReservationId*
 }
 
 type UpdateInfo is {
-	name: String?,
+  name: String?,
+  name: String?,
```
JamesYopp: two name definitions?

src/main/riddl/Events/Reservations/reservationTypeDefinitions.riddl
AlexWeinstein92: These are `Reservation` definitions - there is a commented out `location` field in `Event`, I can simply uncomment it if desired. Probably left it commented as a way of referencing question marks around whether to put it on `Reservation` or `Event`

src/main/riddl/KalixStudy.riddl
AlexWeinstein92: I moved the definitions to their respective `Context.riddl` files, outside the context definition. It was done already for `Member` and I followed the pattern because it seemed better

src/main/riddl/LocationsEventsProjections/locationsEventsProjections.riddl
AlexWeinstein92: That makes sense, I don't think I thought this logic through entirely. Thanks!

I think it might make sense to keep it on Reservation, but I can see it inheriting privacy from Event. That would make sense to me.

src/main/riddl/LocationsEventsProjections/locationsEventsProjections.riddl
AlexWeinstein92: You're totally right, all creation happens from gateway after returning the potentially adjusted event request as a Success case

src/main/riddl/Events/Reservations/reservationTypeDefinitions.riddl
JamesYopp: I the Reservation is the right place for location.  It is, effectively, the join from Event to Location. It would be very convenient to have location with the Event. I think that means we need to work on a projection definition.

AlexWeinstein92: This begs some questions surrounding the issue of whether an Event can have more than one Reservation. 2 ways for that to happen...

One would be cancelled Reservations, but we don't have a ReservationStatus `Cancelled` - maybe we want one? Makes me wonder if they have to move back to Draft to change their Info? That seems not so great.

2nd, we do have both `Event.expectedStart/End` and `Reservation.range`, but maybe that's kind of redundant and can cause sync issues... I think it would make sense to have Event be time agnostic by removing the `expectedStart/End`. If we do this we can also make the Reservation in the Event be a list to allow something like recurrences without going all the way to the ISO spec for RecurrenceRules.

If we go with either of these ideas, it would probably be best to allow `ReservationStatus` to be specified in the query.

src/main/riddl/Tenant/tenant.riddl
AlexWeinstein92: I included these error cases because I thought BillingAccount, DistributionAccount were required fields for getting out of Draft state. Please correct me if I'm wrong :)

src/main/riddl/LocationsEventsProjections/locationsEventsProjections.riddl
AlexWeinstein92: This projection was built for the EasyButton story - it's supposed to take in EventInfo and spit out Locations (or, as I thought, a new Reservation). It really has nothing to do with Events in terms of fields or results. So it is misnamed - should be LocationsReservationsProjection. Sorry for the confusion.

src/main/riddl/Events/Reservations/reservationTypeDefinitions.riddl
JamesYopp: Some good thinking in here Alex.  Here are a few thoughts in response:

I think that an event can have only 1 active reservation. As I read through the different states of Reservation it makes me think that we need to define them a bit better. For example, what is the difference between Reserved and Active? I realize that I was probably the one that created those states. I don't mind the idea of having a Cancelled state, but the question that I have (and I am questioning all the states for reservation at this point) is what business value would it hold? I will spend some time thinking here.

I think there is a case for different time ranges for reservation and event. For example, you may want your reservation for 4 hours before the event starts and 4 hours afterward for setup/teardown. But the Event has a start/end where the actual event takes place. I had not considered recurrence. This may need some thought and conversation. My initial reaction is to move it out of scope for now.

src/main/riddl/Tenant/tenant.riddl
JamesYopp: BillingAccount and DistributionAccount are out of scope for the Kalix-Study. We will need them to be inserted before release. I think we are discussing two different ideas here. Perhaps a call to discuss would be in order to quickly understand both view points.

PR #28, opened 2023-01-18
IA-88: cleaning up organizationsMembersEventsProjections & tenantProjections

AlexWeinstein92:


src/main/riddl/Tenant/tenantMessages.riddl
```diff
@@ -65,15 +65,15 @@ event TenantNameChanged is { tenantId: TenantId, oldName: String, newName: Strin
...
 result TenantListByMember is {
   member: MemberId,
-  organizations: TenantId*
+  tenants: TenantId*
 }
-query FindTenantsByOrganization is {owner: MemberId} briefly "Get a list of organizations that are owned by a member."
-result TenantListByOrganization is {
-  organizations: OrganizationId,
-  owner: TenantId*
+query FindTenantsByOwner is {owner: MemberId} briefly "Get a list of tenants that are owned by a member."
```
JamesYopp: I think it will be rare that a member will own multiple tenants. Actually, as I look at the definition of Tenant there is a PrimaryContact (not a Member), and there is not a concept of Owner. So I am not sure what this query is about. As I look at the deleted lines it may be something I had put in there, and if it was me I am not sure what I was thinking at the time. For now, at least, I don't think we need this query.

src/main/riddl/Tenant/tenantProjections.riddl
```diff
@@ -1,31 +1,27 @@
...
     }
     handler TenantViewsHandler is {
-        on event TenantEstablished {???
-            //then set ^^.TenantViews.orgId to @TenantContext.TenantEstablished.orgId
-            //and set ^^.TenantViews.primaryContact to @TenantContext.TenantEstablished.info.primaryContact
-            //and set ^^.TenantViews.members to @TenantContext.TenantEstablished.info.members
-            //and set ^^.TenantViews.organizations to @TenantContext.TenantEstablished.info.organizations
+        on event TenantEstablished {
+            then "set tenant to TenantEstablished.tenantId"
+            and "set member to TenantEstablished.info.members"
```
JamesYopp: I am debating whether Tenant should have Members. I think this may be a leftover from the idea of a Root Organization (which tenant evolved from). Right now, I am leaning to no.

src/main/riddl/KalixStudy.riddl
```diff
@@ -15,8 +15,10 @@ domain ImprovingApp by author James is {
     include "Product/product"
     include "Order/Order"
     include "Events/eventReservationProjections"
+    include "Tenant/tenantProjections"
     include "LocationsReservationsProjections/locationsReservationsProjections"
-    include "OrganizationsMembersEventsProjections/organizationsMembersEventsProjections"
+    include "OrganizationsForMembersAttendingEventsProjection/organizationsForMembersAttendingEventsProjection"
+    include "MembersAttendingEventsForAnOrganizationProjection/membersAttendingEventsForAnOrganizationProjection"
```
JamesYopp, line 21: I like these names much better. Thanks.

src/main/riddl/Tenant/tenantProjections.riddl
```diff
@@ -1,31 +1,29 @@
+query FindTenantByName is {name: String} briefly "Get a list of tenants that a member belongs to."
```
JamesYopp, line 1: The briefly comment needs to be updated. Maybe, "Get tenant info by searching for it by name."?

JamesYopp (approved):
Thanks Alex

PR #30, opened 2023-01-27
feat: test gateway all services creation [IA-109]

AlexWeinstein92:


src/main/riddl/Test/messages.riddl
```diff
@@ -1,40 +1,116 @@
...
-
 command CreateTenants is {
-  tenantInfos: TestGateway.TenantInfo+
+  tenantInfos: TenantContext.Info+
 } briefly "A command to create multiple tenants at once" explained by {
-| This command will create one tenant for each of the TenantInfo values
-| provided. The TenantInfo has only basic contact information. Members and
-| organizations can be added in batches by another command.
+| This command will create one tenant for each of the Info values
+| provided. The Info has only basic contact information. Organizations can be added in batches by another command.
```
JamesYopp: Given other changes I would drop the last sentence - "Organizations can be added in batches by another command."

src/main/riddl/Test/messages.riddl
```diff
@@ -1,40 +1,116 @@
...
+| provided. The Info includes a lineItems field which can be used to reference product SKUs.
+}
+
+event OrdersCreated is {
+  orders: OrderId+
+} briefly "A reply from creating multiple orders at once" explained by {
+| This event will contain all OrderIds created for the CreateOrders command it replies to.
+}
+
+command CreateOrdersForMembers is {
```
JamesYopp: For now, I think all orders must belong to a Member. At some point, we will want to enable guest orders. But just to speed things up let's just say all orders must be placed by a member. Therefore I don't think we need a CreateOrdersForMembers command. CreateOrders will be sufficient.

src/main/riddl/Test/messages.riddl
```diff
@@ -1,40 +1,116 @@
...
 command CreateTenants is {
-  tenantInfos: TestGateway.TenantInfo+
+  tenantInfos: TenantContext.Info+
 } briefly "A command to create multiple tenants at once" explained by {
-| This command will create one tenant for each of the TenantInfo values
-| provided. The TenantInfo has only basic contact information. Members and
-| organizations can be added in batches by another command.
+| This command will create one tenant for each of the Info values
+| provided. The Info has only basic contact information. Organizations can be added in batches by another command.
 }
```
JamesYopp, line 6: I understand why you are creating these bulk operation commands. It simplifies the creation of the test driver. I think in practice, however, we would want the backend to operate in the singular. The Gateway would provide convenience commands for bulk operations. This would be a great conversation to have on the #improving-App channel.

src/main/riddl/Test/messages.riddl
AlexWeinstein92: The distinction here is that we are specifying members for events, not just using `creatingMember` for all events. Are you saying you would prefer we make individual commands for each member we want to create events for?

src/main/riddl/Test/messages.riddl
AlexWeinstein92: These messages are for the TestGateway and are already intended to be used by it to create entity objects with singular operations

src/main/riddl/Product/product.riddl
```diff
@@ -17,7 +17,6 @@ context ProductContext is {
     event: EventId
   }
   type ProductInfo is {
-    sku: SKU,
```
JamesYopp, line 20: SKU is the productId. I am assuming that you are pulling it from ProductInfo and calling out SKU as an attribute in each of the state definitions? But I don't see that change. So where does SKU go?

src/main/riddl/Store/Store.riddl
```diff
@@ -3,7 +3,6 @@ context StoreContext is {
   type StoreId is Id(StoreContext.Store)
 
   type StoreInfo is {
-    id: StoreId,
```
JamesYopp, line 6: I am assuming that you are pulling storeId from StoreInfo and calling out storeId as an attribute in each of the state definitions? But I don't see that change. So where does StoreId go?

src/main/riddl/Product/product.riddl
AlexWeinstein92: Any EntityID can be found using `context.entityId` once you are inside an entity, so it's redundant to have IDs in state. I can include it in Product entity like we do in the other entities, or remove from everywhere

src/main/riddl/Test/messages.riddl
JamesYopp: Perfect

src/main/riddl/Test/messages.riddl
JamesYopp: CreateMember is one command. Once a Member is created they can place an Order. You have a pair of commands - CreateOrder, and CreateOrdersForMembers. I don't see the reason for both. You have to have a valid member to create an order. The alternate is a Guest order, which would not require a member to place an order. So I am suggesting that we just go with CreateOrder for now (because it implies you have a valid Member). Later we will introduce CreateGuestOrder. I think the intent of a RIDDL mapping is zipping together two lists. I think you are looking at this like a map command where we have a list of Members and an (singular) order template. Using the template we create an order for each of the members in the list. Is that the intent?

JamesYopp (approved):

PR #31, opened 2023-01-30
feat: runScenario projections [IA-108]

AlexWeinstein92:


src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -0,0 +1,103 @@
...
+  projection AllTenantIdsView is {
+     fields {
+        tenant: TenantId
+    }
+    handler AllTenantIdsViewHandler is {
+        on event TenantEstablished {
+            then "create row with tenant set to TenantEstablished.tenantId"
+        }
+        on query GetAllTenantIds {
+            then "reply with AllTenantIds packed with entire table"
```
JamesYopp: I am not looking at type definitions ATM, but I wonder if it would be worth qualifying this with all "Active" TenantIDs?

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -0,0 +1,103 @@
...
+  projection AllOrgIdsView is {
+     fields {
+        org: OrganizationId
+    }
+    handler AllOrgIdsViewHandler is {
+        on event OrganizationEstablished {
+            then "create row with org set to OrganizationEstablished.orgId"
+        }
+        on query GetAllOrgIds {
+            then "reply with AllOrgIds packed with entire table"
```
JamesYopp: Same as tenant? All Active OrgIds?

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -0,0 +1,103 @@
...
+  projection AllEventIdsView is {
+     fields {
+        event: EventId
+    }
+    handler AllTenantIdsViewHandler is {
+        on event EventScheduled {
+            then "create row with event set to EventEstablished.eventId"
+        }
+        on query GetAllEventIds {
+            then "reply with AllEventIds packed with entire table"
```
JamesYopp: All scheduled events?

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -0,0 +1,103 @@
...
+  projection AllMemberIdsView is {
+     fields {
+        member: MemberId
+    }
+    handler AllMemberIdsViewHandler is {
+        on event MemberRegistered {
+            then "create row with member set to MemberRegistered.memberId"
+        }
+        on query GetAllMemberIds {
+            then "reply with AllMemberIds packed with entire table"
```
JamesYopp, line 101: all active members?

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -0,0 +1,103 @@
...
+  projection AllStoreIdsView is {
+     fields {
+        store: StoreId
+    }
+    handler AllStoreIdsViewHandler is {
+        on event StoreCreated {
+            then "create rows with store set to StoreCreated.storeId"
+        }
+        on query GetAllStoreIds {
+            then "reply with AllStoreIds packed with entire table"
```
JamesYopp, line 115: Again, not looking at states, but all active stores?

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -0,0 +1,103 @@
...
+  projection AllSkusView is {
+     fields {
+        sku: SKU
+    }
+    handler AllSkusViewHandler is {
+        on event ProductCreated {
+            then "create rows with sku set to ProductCreated.sku"
+        }
+        on query GetAllSkus {
+            then "reply with AllSkus packed with entire table"
```
JamesYopp, line 129: all available skus?

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
AlexWeinstein92: Looking at states - it seems there is a `Ready` state on Stores, but there is no way to get into that state from Draft, and you must be in `Ready` state before going into `Open`. We probably want to filter on `Open`.

AlexWeinstein92:
@JamesYopp I am currently implementing your suggestions, but it occurs to me that this will require runScenario to handle state changes to make sure entities are in correct state before querying for these IDs so dependent entities can be created. Which means we probably want batch `UpdateStatus` endpoints in TestGateway to facilitate. Might make things a bit messier, just saying, we don't neccesarily need to be realistic in these projections as they are meant for the Kalix study tests

JamesYopp (approved):

JamesYopp (approved):

PR #32, opened 2023-01-31
feat: TestGateway Handler [IA-110]

AlexWeinstein92:
Should also satisfy IA-104 https://yoppworks.atlassian.net/browse/IA-104


src/main/riddl/Test/messages.riddl
```diff
@@ -5,7 +5,7 @@ command CreateTenants is {
 | provided. The Info has only basic contact information.
 }
 
-event TenantsCreated is {
+result TenantsCreatedResult is {
```
JamesYopp: This confuses me. TenantsCreated is an event that is generated in response to a command, right? Results are the response to a Query.

JamesYopp (approved):

PR #33, opened 2023-01-31
Refactor the fields keyword to 'record' for proper validation.

miguelTaningcoYoppworks:
### Issue

Currently, the Kalix-Study branch does not validate and therefore also fail in generating the static files. It turns out that the "fields" keyword has been changed such that projections require "records".

### Changes

- refactor 'fields' keywords to 'record foo'. eg. AllTenantIdsView will have the record AllTenantIdsFields

### Testing
- validated in local machine
- checked hugo server after translating

### DIFF
--- a/src/main/riddl/AllIdsProjections/allIdsProjections.riddl
+++ b/src/main/riddl/AllIdsProjections/allIdsProjections.riddl
@@ -20,7 +20,7 @@ result AllSkus is {skus: SKU*}
 
 context AllIdsProjections is  {
   projection AllTenantIdsView is {
-     record fields {
+     record AllTenantIdsFields {
         tenant: TenantId
         //tenantStatus: TenantContext.Status
     }
@@ -38,7 +38,7 @@ context AllIdsProjections is  {
   }
 
   projection AllOrgIdsView is {
-     record fields {
+     record AllOrgIdsFields {
         org: OrganizationId
         //orgStatus: OrganizationContext.Status
     }
@@ -56,7 +56,7 @@ context AllIdsProjections is  {
   }
 
   projection AllEventIdsView is {
-     record fields {
+     record AllEventIdsFields{
         event: EventId
         //eventStatus: EventContext.Status
     }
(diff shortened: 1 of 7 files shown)


src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -20,7 +20,7 @@ result AllSkus is {skus: SKU*}
 
 context AllIdsProjections is  {
   projection AllTenantIdsView is {
-     fields {
+     record fields {
```
JamesYopp: I am curious, in most other areas of RIDDL you have a keyword, in this case record, followed by the name of that thing. 

So my question is, did the keyword change to record fields? or is it just record <name>?

In other words, if you changed this line to:
  record foo {
Would it still compile?

If yes, then should we be considering a different or better name for this record definition aside from fields?

reid-spencer: Its just "record".  I changed the title of this PR :)

reid-spencer:
@miguelTaningcoYoppworks - See the comments by James and me above. Your "Validate RIDDL" failed because of a syntax error. The run log also indicates, with this error:
```
Error:  Error: tenantProjections.riddl(9:5): Expected one of ("}" | "invariant" | "handler" | "include" | "term" | "{" | "fields" | "option" | "options" | "???"):
    record fields {
```
That the validator is using an old version of riddlc.  I will have to make a new release. This is a result of very new feature changes.  So we'll leave this sit until the new release can be configured.

reid-spencer:
@miguelTaningcoYoppworks - I have released riddlc 0.18.1 which includes the projection parsing fixes and corrected the get-riddlc action to use it. The "validate-riddl" workflow should now use 0.18.1 and require the use of at least one "record" definition in a projection.  If you push the fix suggested by me in response to James' comment, it should all work! :)

src/main/riddl/AllIdsProjections/allIdsProjections.riddl
miguelTaningcoYoppworks: My bad for the vagueness. Thanks for updating the title! I've changed the 'fields' names to be more descriptive by having the View name and replacing View with Fields.

eg. AllTenantIdsView will have the record AllTenantIdsFields

JamesYopp (approved):
Looks good. Thanks Miguel!

PR #34, opened 2023-02-15
Update Order RIDDL based on kalix-study codebase

miguelTaningcoYoppworks:


src/main/riddl/Order/Order.riddl
```diff
@@ -4,16 +4,16 @@ context OrderContext is {
   type LineItem is {
     product: SKU,
     quantity: Number,
-    unitPrice: Number,
```
JamesYopp, line 7: I think that maybe you have this a bit backward. unitPrice is the price of quantity 1. lineTotal = quantity * unitPrice. If we were to delete one of the two attributes I would delete lineTotal because it is a calculated attribute.

src/main/riddl/Order/Order.riddl
```diff
@@ -4,16 +4,16 @@ context OrderContext is {
...
     lineTotal: Number
   }
 
   type OrderInfo is {
+    order_id: OrderId,
     lineItems: LineItem+,
-    specialInstructions: String,
```
JamesYopp, line 13: We probably don't need specialInstructions for the KalixStudy. However, I think we need it for future iterations. So rather than delete it, what if we made it optional? 
`specialInstructions: String?`

src/main/riddl/Order/Order.riddl
```diff
@@ -4,16 +4,16 @@ context OrderContext is {
...
-    orderTotal: Number
```
JamesYopp, line 14: Yup. good with this change. But we have to remember that we need to calculate the line total for each line item and then sum all the lineTotals from each lineItem in lineItems.

src/main/riddl/Order/Order.riddl
```diff
@@ -4,16 +4,16 @@ context OrderContext is {
...
+    specialInstructions: String
```
JamesYopp: I see. You just moved this attribute. Still, I think special instructions should be optional
specialInstructions: String?

src/main/riddl/Order/Order.riddl
```diff
@@ -35,7 +34,7 @@ context OrderContext is {
   command CancelOrder {orderId: OrderId, cancellingMember: MemberId}
   event OrderCancelled {orderId: OrderId, info: OrderInfo, meta: OrderMetaInfo}
   query GetOrderInfo {orderId: OrderId, requstingMembert: MemberId}
-  result OrderInfoResult {orderId: OrderId, info: OrderMetaInfo}
+  result OrderInfoResult {orderId: OrderId, info: OrderInfo}
```
JamesYopp, line 54: To get a complete picture of the order we may want both OrderInfo and OrderMetaInfo
result OrderInfoResult {orderId: OrderId, info: OrderInfo, meta: OrderMetaInfo}

src/main/riddl/Order/Order.riddl
```diff
@@ -109,7 +108,7 @@ context OrderContext is {
             and "yield OrderStatusUpdated"
           }
           example PendingOrderIllegalStateTransition is {
-            given "order is transitioning from Pending to Draft, Ready, or Delivered"
+            given "order is transitioning from Pending to Pending, Draft, Ready, or Delivered"
```
JamesYopp: I don't think it makes sense for order to transition from Pending to Pending...

src/main/riddl/Order/Order.riddl
```diff
@@ -141,13 +140,13 @@ context OrderContext is {
             and "yield OrderStatusUpdated event"
           }
           example InProcessOrderIllegalStateTransition is {
-            given "order is transitioning from InProcess to Draft, Pending, Cancelled, or Delivered"
+            given "order is transitioning from InProcess to InProcess, Draft, Pending, Cancelled, or Delivered"
```
JamesYopp: Can somebody cancel an InProcess order? I think they should be able to. I think the rest is valid.

src/main/riddl/Order/Order.riddl
```diff
@@ -141,13 +140,13 @@ context OrderContext is {
...
             then error "Draft orders may not transition to desired state."
           }
         }
-        on command UpdateOrderInfo {then error "InProcess Orders may not be modified"}
+        on command UpdateOrderInfo {then error "InProcess Orders may not be modified."}
         on command CancelOrder {then error "InProcess Orders may not be cancelled."}
-        on query GetOrderInfo {then "yeild OrderInfoResult message"}
+        on query GetOrderInfo {then "yield OrderInfoResult message"}
```
JamesYopp, line 161: Either I copy/pasted a lot or I just can't spell yield... Either way, thanks for cleaning up!

src/main/riddl/Order/Order.riddl
```diff
@@ -194,10 +198,15 @@ context OrderContext is {
...
-        on query GetOrderInfo {???}
+        on command UpdateOrderStatus {
+          example ReadyOrderIllegalStateTransition is {
+            given "order is transitioning from Cancelled to Draft, Pending, Cancelled, InProcess, or Ready, Delivered"
+            then error "Cancelled orders may not transition to desired state."
+          }
+        }
+        on command UpdateOrderInfo {then error "Cancelled Orders may not be modified."}
+        on command CancelOrder {then error "Delivered Orders may not be cancelled."}
+        on query GetOrderInfo {then "yield OrderInfoResult message"}
```
JamesYopp: on query GetOrderInfo {then "yield result OrderInfoResult"}

JamesYopp (commented):
Thanks Miguel. See comments to changes. Happy to talk through them if you would like.

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Ahh interesting. Currently in the code base, the order has an orderTotal attribute. This is a calculated attribute in which orderTotal = summation of lineTotal * quantity per line item. Is this wrong?

should it be orderTotal = summation of lineTotal per line item, in which each lineTotal = quantity * unitPrice?

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: sounds good!

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: sounds good. Will apply the appropriate changes once the above comment is addressed

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: I think when Alex and I were talking, it was that an InProcess order would be very hard to cancel. So for that reason, we agreed that an InProcess order can't cancel

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Yes it should ^ :p

src/main/riddl/Order/Order.riddl
reid-spencer: @miguelTaningcoYoppworks - please revert

src/main/riddl/Order/Order.riddl
reid-spencer: @miguelTaningcoYoppworks - which comment above?

src/main/riddl/Order/Order.riddl
reid-spencer: @miguelTaningcoYoppworks - has this been done?

src/main/riddl/Order/Order.riddl
reid-spencer: @miguelTaningcoYoppworks - is this fixed yet?

src/main/riddl/Order/Order.riddl
reid-spencer: @improving-app/developers - Generally it is a bad idea to make inconsistencies for the user that are state dependent. They should be able to cancel and order at any time. If the order is "In Process" then the system needs to do what it needs to do to cancel that order, even if it is hard, like recalling an order that is being delivered. In our scenario, there isn't a huge delay between OrderReady and OrderDelivered because of the locality of events to customers.  However, delivery persons could carry a cell phone that would get an instant message to return the item to the store, or other strategies.

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Left a direct message to Reid asking how calculated fields must be documented on riddl or how fields only present internally must be documented.

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: The comment was relating to "I think that maybe you have this a bit backward. unitPrice is the price of quantity 1. lineTotal = quantity * unitPrice. If we were to delete one of the two attributes I would delete lineTotal because it is a calculated attribute."

This conversation can be closed since it's really relating to the same thing.

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: 👍 Will go ahead and allow to cancel an in-process order. Will also add a separate ticket for reflecting thing back to the improving-app repository

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Updated branch to make this optional ^

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Reverted 👍

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Updated PR to allow InProcess -> Cancelled

src/main/riddl/Order/Order.riddl
```diff
@@ -23,8 +23,7 @@ context OrderContext is {
   }
```
miguelTaningcoYoppworks, line 31: At line 3 and line 22 are references to the OrderStatus type. @reid-spencer as per our talks yesterday on this topic, can you confirm that we want to remove something like this? If so, @JamesYopp I just wanted to let you know about this :)

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: For now we are just going to add some documentation to say these are calculated

src/main/riddl/Order/Order.riddl
miguelTaningcoYoppworks: Removed everything status related from order.riddl assuming this is preferred.

reid-spencer: Okay.

reid-spencer (dismissed):
LGTM

reid-spencer (approved):
LGTM

PR #35, opened 2023-02-15
Update Member RIDDL based on kalix-study codebase

miguelTaningcoYoppworks:


src/main/riddl/Members/members.riddl
```diff
@@ -12,15 +12,15 @@ context MembersContext  is {
...
     type Info is {
         contact: Contact,
         handle: String briefly "username. Must be unique.",
         avatar: URL briefly "thumbnail image of the user. Need to screen to ensure community guidelines compliance. i.e. no porn or offensive content. Use Gravitar?",
         firstName: String briefly "Real first name of the member",
         lastName: String briefly "Real last name of the member",
-        mobileNumber: MobileNumber? briefly "Optional mobile phone number",
-        emailAddress: EmailAddress? briefly "Optional email address",
+        //mobileNumber: MobileNumber? briefly "Optional mobile phone number",
+        //emailAddress: EmailAddress? briefly "Optional email address",
```
JamesYopp, line 26: I am ok dropping email and mobile for the KalixStudy. Thanks for commenting so we don't lose the idea.

src/main/riddl/Members/members.riddl
```diff
@@ -127,6 +127,60 @@ context MembersContext  is {
...
+        state Draft is {
+          fields {
+            info: MembersContext.Info,
+            meta: MembersContext.MetaInfo
+          }
+
+          handler DraftMemberHandler {
+            on command RegisterMember {then error "Member has already been registered."}
+            on command UpdateMemberStatus {
+              example DraftMemberToActiveMember is {
```
JamesYopp: Alex and I have had several conversations about Members registering themselves. I think it is a common use case. I don't think the riddl needs to change here, but we should be explicit that the RegisteringMember may be "self" or the entity's own ID.

src/main/riddl/Members/members.riddl
```diff
@@ -127,6 +127,60 @@ context MembersContext  is {
...
+                given "entity is transitioning from Draft to Active and required fields are non-empty"
+                then "set entity Status to Status field from message"
+                and "set entity lastModifiedBy to actingMember from message"
+                and "set entity lastModifiedOn to today()"
+                and "yield event MemberStatusUpdated"
+              }
+              example DraftMemberToInactiveMember is {
```
JamesYopp: I think the only state transitions for Member from the Draft state are Draft->Active and Draft->Terminated. I think all others should throw an error.

src/main/riddl/Members/members.riddl
```diff
@@ -127,6 +127,60 @@ context MembersContext  is {
...
+                given "entity is transitioning from Draft to Inactive and required fields are non-empty"
+                then "set entity Status to Status field from message"
+                and "set entity lastModifiedBy to actingMember from message"
+                and "set entity lastModifiedOn to today()"
+                and "yield event MemberStatusUpdated"
+              }
+              example DraftMemberToSuspendedMember is {
```
JamesYopp: I think the only state transitions for Member from the Draft state are Draft->Active and Draft->Terminated. I think all others should throw an error.

src/main/riddl/Members/members.riddl
```diff
@@ -127,6 +127,60 @@ context MembersContext  is {
...
+              }
+            }
+            on command UpdateMemberInfo {
+              then "set entity Info to non-Empty fields in UpdateInfo from message"
+              and "set entity lastModifiedBy to actingMember from message"
+              and "set entity lastModifiedOn to today()"
+              and "yield event MemberInfoUpdated"
+            }
+            on query GetMemberData {
+              then "yield Info for entity"
```
JamesYopp, line 271: eventually we will need to worry about permissions here. I think only admins and self should be able to make this call. Or, maybe other roles can get a limited set of info (in which case we would need another aggregate?). For the KalixStudy the way you have it is fine. Just notes for future reference.

src/main/riddl/Members/members.riddl
```diff
@@ -157,24 +211,126 @@ context MembersContext  is {
...
+            }
+          }
+        }
+
+        state Suspended is {
+          fields {
```
JamesYopp: I think we should add:
reason: String
as a required field for the suspended state. In other words, you cannot suspend a Member without explaining the action. So when someone attempts to reactivate the member they can determine if there has been proper resolution to the reason for suspension.

src/main/riddl/Members/members.riddl
```diff
@@ -157,24 +211,126 @@ context MembersContext  is {
...
+                and "set entity lastModifiedBy to actingMember from message"
+                and "set entity lastModifiedOn to today()"
+                and "yield event MemberStatusUpdated"
+              }
+              example SuspendedMemberIllegalStateTransition is {
+                given "entity is transitioning from Suspended to Draft, Inactive, or Suspended"
+                then error "Draft members may not transition to desired state."
+              }
+            }
+            on command UpdateMemberInfo {
```
JamesYopp, line 416: Do we also need to add an UpdateSuspendedReason command - or have a different aggregate (SuspendedMemberInfo?) to contain reason?

JamesYopp (commented):
See comments. Happy to talk them through if you would like.

src/main/riddl/Members/members.riddl
miguelTaningcoYoppworks: I will add this as a comment in riddl to make sure this is logged.

src/main/riddl/Members/members.riddl
miguelTaningcoYoppworks: Maybe MetaInfo should have an optional suspensionReason, and UpdateMemberStatus to have reason: String, where it is optional, but required for updating to the suspended state? These changes should allow us to have what you proposed

src/main/riddl/Members/members.riddl
reid-spencer: @miguelTaningcoYoppworks - probably all that is needed to address James' comment is some documentation on how the RegisterMember message definition

src/main/riddl/Members/members.riddl
reid-spencer: The way James has modeled these entities is as state machines. Suspended is a state. The "reason" only needs to exist when the entity is in the Suspended state. So, I think UpdatedSuspensionReason is a reasonable command, but the Suspended state could also just accept another SuspendMember command and update the "reason" status as part of that.

miguelTaningcoYoppworks: Alright I think that makes sense. I will add UpdateSuspensionReason and UpdatedSuspensionReason commands. As for storing the "reason" field, should this be added into the MetaInfo? I'm not too sure where this should be persisted.

src/main/riddl/Members/members.riddl
miguelTaningcoYoppworks: Will add documentation! Out of curiosity, does that mean that the RegisterMember command's registeringMember field must now be optional? I'm not too sure how it works if the member is registering themselves and couldn't fill in the registeringMember field. Or does it not work like that?

reid-spencer: No, not optional. It means that the value for regtieringMember is the same Id(Member)  as the member being registered. And, that's okay.  The documentation should say as much.

src/main/riddl/Members/members.riddl
reid-spencer: I would rather you just allowed the SuspendMember message to be handled when in the Suspended state.

reid-spencer: The "reason" field should be called "suspensionReason" and be part of the Suspended state's fields.

src/main/riddl/Members/members.riddl
miguelTaningcoYoppworks: Added explicitly in RegisterMember command ^

src/main/riddl/Members/members.riddl
miguelTaningcoYoppworks: Alright, so to summarize:

- Add suspensionReason: String to the Suspended state
- Add SuspendMember command which has the memberId and suspensionReason (should error except in the Suspended state)
- Add SuspendedMember event which has the memberId and newSuspensionReason

src/main/riddl/Members/members.riddl
reid-spencer: Yes, that's correct.

src/main/riddl/Members/members.riddl
reid-spencer: The middle bullet is incorrect. It should be allowed in all states.  Whether you've got an Active or Draft state for member, they should be suspendable.

miguelTaningcoYoppworks: As per our messages offline, here are the changes applied:

- Removed Status in members.riddl. 
- Removed memberStatus. 
- Removed UpdateMemberStatus. 
- Removed MemberStatusUpdated. 
- Add ActivateMember, MemberActivated, InactivateMember, MemberInactivated, SuspendMember, MemberSuspended, TerminateMember, MemberTerminated. 
- Add suspensionReason to the Suspended state.

reid-spencer (dismissed):
LGTM

reid-spencer (approved):
LGTM

PR #44, opened 2023-03-20
feat: include KalixStudy dependencies directly in ImprovingApp [IA-132]

AlexWeinstein92:


src/main/riddl/Commerce/commerce.riddl
```diff
@@ -8,23 +8,6 @@ context Commerce is {
...
-    type OrderInfo is {
-        customerId: MemberId, //TODO: This should be CustomerId but not sure how to model that in RIDDL
-        lineItems: LineItem+,
-        itemTotal: Number,
-        taxTotal: Number,
-        deliveryTotal: Number, //TODO: Should delivery just be a LineItem?
-        grandTotal: Number,
-        paymentId: PaymentId
-    }
     type DeliveryId is Id( Delivery )
```
reid-spencer, line 11: Why are we removing these? It doesn't look like you moved them anywhere, as no other file includes this information.  How is an Order supposed to be represented in this application?

reid-spencer (commented):
Mostly okay, but I don't like the large scale deletion of Order, OrderInfo and LineItem

src/main/riddl/Commerce/commerce.riddl
AlexWeinstein92: These are already modelled in Order.riddl from Kalix Study. I talked it over with James and he agreed to take them out of Commerce because Order is big enough for its own entity now. Order was previously not included in ImprovingApp domain

reid-spencer: Okay good enough for me.

PR #45, opened 2023-03-21
IA 128 status and state

AlexWeinstein92:
See https://yoppworks.atlassian.net/browse/IA-128

### DIFF
--- a/src/main/riddl/Members/members.riddl
+++ b/src/main/riddl/Members/members.riddl
@@ -5,6 +5,7 @@ type MemberMap is mapping from MemberId to MembersContext.Info
 
 context MembersContext  is {
     //Member Details
+    type StatusUpdates is any of {Active, Inactive, Suspended} briefly "Internal status of an Organization."
 
     type NotificationPreference is any of { Email, SMS, Application } described
     as {
@@ -146,30 +147,14 @@ context MembersContext  is {
       |that the member is now active.
     }
 
-    command InactivateMember is {memberId: MemberId, actingMember: MemberId}
-    briefly "Update state of member to Inactive"
-    described as {
-      |This command inactivates the member.
+    command UpdateMemberStatus is {
+      memberId: MemberId,
+      status: ^MemberContext.StatusUpdates,
+      updatingMember: MemberId
     }
-
-    event MemberInactivated is {memberId: MemberId, meta: MembersContext.MetaInfo}
-    briefly "Event issued after InactivateMember"
-    described as {
-      |This event is issued after InactivateMember to denote
-      |that the member is now inactive.
-    }
-
-    command SuspendMember is {memberId: MemberId, actingMember: MemberId, suspensionReason: String}
-    briefly "Update state of member to Suspended"
-    described as {
-      |This command suspends the member.
-    }
-
-    event MemberSuspended is {memberId: MemberId, meta: MembersContext.MetaInfo}
-    briefly "Event issued after SuspendMember"
-    described as {
-      |This event is issued after SuspendMember to denote
-      |that the member is now suspended.
+    event MemberStatusUpdated is {
+      memberId: MemberId,
+      newStatus: ^MemberContext.StatusUpdates
     }
 
     command TerminateMember is {memberId: MemberId, actingMember: MemberId}
@@ -245,12 +230,6 @@ context MembersContext  is {
               and "set entity lastModifiedOn to today()"
               and "yield event MemberActivated"
             }
-            on command InactivateMember {
-              then error "Cannot set state to Inactive on a member that is not active"
-            }
-            on command SuspendMember {
-              then error "Cannot set state to Suspended on a member that is not active"
-            }
             on command TerminateMember {
               given "entity has non-empty required fields"
               then "morph Member entity to Terminated state"
@@ -277,19 +256,19 @@ context MembersContext  is {
           |back to Draft.
         }
 
-        record ActiveState is {
+        record ActiveMemberState is {
             info: MembersContext.Info,
             meta: MembersContext.MetaInfo
         }
-        state Active of ^ActiveState is {
-            invariant time_sanity is {
-              <=(@Active.meta.createdOn, @Active.meta.lastModifiedOn)
-            }
+        state ActiveMember of ActiveMemberState is {
+            //invariant time_sanity is {
+            //  <=(@ActiveMemberState.meta.createdOn, @ActiveMemberState.meta.lastModifiedOn)
+            //}
             invariant firstName is {
-              >=(length(@Active.info.firstName), 3)
+              >=(length(@ActiveMemberState.info.firstName), 3)
             }
             invariant lastName is {
-              >=(length(@Active.info.firstName), 3)
+              >=(length(@ActiveMemberState.info.firstName), 3)
             }
 
 /*
@@ -313,21 +292,13 @@ context MembersContext  is {
                     and "set entity lastModifiedOn to today()"
                     and "yield event MemberInfoUpdated"
                 }
-                on command ActivateMember {
-                  then error "Cannot activate a member that is already activated"
-                }
-                on command InactivateMember {
-                  then "morph Member entity to Inactive state"
-                }
-                on command SuspendMember {
-                  given "entity has suspensionReason non-empty"
-                  then "morph Member entity to Suspended state"
-                  and "set entity lastModifiedBy to actingMember from message"
-                  and "set entity lastModifiedOn to today()"
-                  and "yield event MemberSuspended"
+                on command UpdateMemberStatus {
+                    then "morph entity Member to state provided in UpdateMemberStatus"
+                    and "set Meta data to reflect the change"
+                    and "yield event MemberStatusUpdated"
                 }
                 on command TerminateMember {
-                  then "morph Member entity to Terminated state"
+                    then morph entity Member to state MemberTerminated with @ActiveMemberState.meta
                 }
                 on query GetMemberData {
                     then "yield Info for entity"
@@ -338,103 +309,10 @@ context MembersContext  is {
           |member can transition to any state other than Draft.
         }
 
-        record InactiveState is {
-          info: MembersContext.Info,
-          meta: MembersContext.MetaInfo
+        record TerminatedMemberState is {
+            meta: MembersContext.MetaInfo
         }
-        state Inactive of ^InactiveState is {
-          handler InactiveMemberHandler {
-            on command RegisterMember {then error "Member has already been registered."}
-            on command ActivateMember {
-              given "entity has non-empty required fields"
-              then "morph Member entity to Active state"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberActivated"
-            }
-            on command InactivateMember {
-              then error "Cannot set a member to inactive that is already inactivated"
-            }
-            on command SuspendMember {
-              then error "Cannot set an inactive member to suspended"
-            }
-            on command TerminateMember {
-              given "entity has non-empty required fields"
-              then "morph Member entity to Terminated state"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberTerminated"
-            }
-            on command UpdateMemberInfo {
-              then "set entity Info to non-Empty fields in UpdateInfo from message"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberInfoUpdated"
-            }
-            on query GetMemberData {
-              then "yield Info for entity"
-            }
-          }
-        } described as {
-          |This state describes that the member will no longer be active.
-          |One example is when a student graduates but the admin wants to
-          |keep the record.
-        }
-
-        record SuspendedState is {
-          info: MembersContext.Info,
-          meta: MembersContext.MetaInfo,
-          suspensionReason: String
-        }
-        state Suspended of ^SuspendedState is {
-          handler SuspendedMemberHandler {
-            on command RegisterMember {then error "Member has already been registered."}
-            on command ActivateMember {
-              given "entity has non-empty required fields"
-              then "morph Member entity to Active state"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberActivated"
-            }
-            on command InactivateMember {
-              then error "Cannot set a suspended member to inactive"
-            }
-            on command SuspendMember {
-              given "entity has suspensionReason non-empty"
-              then "morph Member entity to Active state"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberSuspended"
-            } briefly "A suspension reason can be updated by re-issuing the command"
-            on command TerminateMember {
-              given "entity has non-empty required fields"
-              then "morph Member entity to Terminated state"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberTerminated"
-            }
-            on command UpdateMemberInfo {
-              then "set entity Info to non-Empty fields in UpdateInfo from message"
-              and "set entity lastModifiedBy to actingMember from message"
-              and "set entity lastModifiedOn to today()"
-              and "yield event MemberInfoUpdated"
-            }
-            on query GetMemberData {
-              then "yield Info for entity"
-            }
-          }
-        } described as {
-          |This state describes that a member cannot access anything due
-          |to being a bad actor. This state transition is only possible
-          |as a command issued by an administrator and therefore can only
-          |be removed from the suspended state by an administrator.
-        }
-
-
-        record TerminatedState is {
-            memberId: MemberId // needed for compilation - RIDDL requires fields object and fields object requires at least one field
-        }
-        state Terminated of ^TerminatedState is {
+        state MemberTerminated of TerminatedMemberState is {
             handler TerminatedMemberHandler is {
                 on other { then error "Terminated members cannot process messages" }
             }
--- a/src/main/riddl/Members/members.riddl
+++ b/src/main/riddl/Members/members.riddl
@@ -5,13 +5,41 @@ type MemberMap is mapping from MemberId to MembersContext.Info
 
 context MembersContext  is {
     //Member Details
-    type StatusUpdates is any of {Active, Inactive, Suspended} briefly "Internal status of an Organization."
+    type MemberStatuses is any of {Active, Suspended} briefly "Internal status of an Organization."
 
     type NotificationPreference is any of { Email, SMS, Application } described
     as {
       | The values of this enumeration define the ways in which a member may
       | be notified of changes.
     }
+
+    type RequiredDraftInfo is {
+        contact: Contact
+          briefly "Contact information of the member",
+        handle: String
+          briefly "username. Must be unique.",
+        avatar: URL
+          briefly "thumbnail image of the user. Need to screen to ensure community guidelines compliance. i.e. no porn or offensive content. Use Gravitar?",
+        firstName: String
+          briefly "Real first name of the member",
+        lastName: String
+          briefly "Real last name of the member",
+        tenant: TenantId
+          briefly "Tenant id of the member"
+    } described as {
+      |This type provides the basic information about members that is
+      |required for transitioning to Established state. The fields of this type form the profile
+      |of the Member with demographics, contact info, and various options
+      |like notifications and group membership the Member can opt in to.
+    }
+
+    type OptionalDraftInfo is {
+        notificationPreference: NotificationPreference?
+            briefly "if this is set it is assumed that the user has opted in.",
+        organizationMembership: OrganizationId*
+            briefly "A possibly empty list of organizations this member belongs to."
+    }
+
     type Info is {
         contact: Contact
           briefly "Contact information of the member",
@@ -37,7 +65,7 @@ context MembersContext  is {
       |of the Member with demographics, contact info, and various options
       |like notifications and group membership the Member can opt in to.
     }
-    type OptionalInfo is {
+    type EditInfo is {
         handle: String?
           briefly "username. Must be unique.",
         avatar: URL?
@@ -118,8 +146,8 @@ context MembersContext  is {
     //  registeringMember: MemberId
     //}
 
-    command UpdateMemberInfo is {memberId: MemberId, info: MembersContext.OptionalInfo, actingMember: MemberId}
-    briefly "Update info of a member"
+    command EditMemberInfo is {memberId: MemberId, info: MembersContext.EditInfo, actingMember: MemberId}
+    briefly "Edit info of a member"
     described as {
       |This command will overwrite the values of attributes provided in UpdateInfo.
       |If an attribute is not referenced it will be left as is. If an attribute is
@@ -128,10 +156,10 @@ context MembersContext  is {
       |overwrite what is stored.
     }
 
-    event MemberInfoUpdated is {memberId: MemberId, info: MembersContext.Info, meta: MembersContext.MetaInfo}
-    briefly "Event issued after UpdateMemberInfo"
+    event MemberInfoEdited is {memberId: MemberId, info: MembersContext.Info, meta: MembersContext.MetaInfo}
+    briefly "Event issued after EditMemberInfo"
     described as {
-      |This event occurs whenever a member's information is updated.
+      |This event occurs whenever a member's information is edited.
     }
 
     command ActivateMember is {memberId: MemberId, actingMember: MemberId}
@@ -147,14 +175,18 @@ context MembersContext  is {
       |that the member is now active.
     }
 
-    command UpdateMemberStatus is {
-      memberId: MemberId,
-      status: ^MemberContext.StatusUpdates,
-      updatingMember: MemberId
+
+    command SuspendMember is {memberId: MemberId, suspendingMember: MemberId}
+    briefly "Update state of member to Suspended"
+    described as {
+      |This command suspends the member such that most operations are not available.
     }
-    event MemberStatusUpdated is {
-      memberId: MemberId,
-      newStatus: ^MemberContext.StatusUpdates
+
+    event MemberSuspended is {memberId: MemberId, meta: MembersContext.MetaInfo}
+    briefly "Event issued after SuspendMember"
+    described as {
+      |This event is issued after SuspendMember to denote
+      |that the member is now suspended.
     }
 
     command TerminateMember is {memberId: MemberId, actingMember: MemberId}
@@ -216,11 +248,12 @@ context MembersContext  is {
             //}
         }
 
-        record DraftState is {
-          info: MembersContext.OptionalInfo,
+        record DraftMemberState is {
+          requiredInfo: MembersContext.RequiredDraftInfo,
+          optionalInfo: MembersContext.OptionalDraftInfo,
           meta: MembersContext.MetaInfo
         }
-        state Draft of ^DraftState is {
+        state DraftMember of DraftMemberState is {
           handler DraftMemberHandler {
             on command RegisterMember {then error "Member has already been registered."}
             on command ActivateMember {
@@ -237,7 +270,7 @@ context MembersContext  is {
               and "set entity lastModifiedOn to today()"
               and "yield event MemberTerminated"
             }
-            on command UpdateMemberInfo {
+            on command EditMemberInfo {
               then "set entity Info to non-Empty fields in UpdateInfo from message"
               and "set entity lastModifiedBy to actingMember from message"
               and "set entity lastModifiedOn to today()"
@@ -256,19 +289,19 @@ context MembersContext  is {
           |back to Draft.
         }
 
-        record ActiveMemberState is {
+        record RegisteredMemberState is {
             info: MembersContext.Info,
             meta: MembersContext.MetaInfo
         }
-        state ActiveMember of ActiveMemberState is {
-            //invariant time_sanity is {
-            //  <=(@ActiveMemberState.meta.createdOn, @ActiveMemberState.meta.lastModifiedOn)
-            //}
+        state RegisteredMember of RegisteredMemberState is {
+            invariant time_sanity is {
+              <=(@RegisteredMemberState.meta, @RegisteredMemberState.meta)
+            }
             invariant firstName is {
-              >=(length(@ActiveMemberState.info.firstName), 3)
+              >=(length(@RegisteredMemberState.info.firstName), 3)
             }
             invariant lastName is {
-              >=(length(@ActiveMemberState.info.firstName), 3)
+              >=(length(@RegisteredMemberState.info.firstName), 3)
             }
 
 /*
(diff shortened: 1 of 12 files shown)


src/main/riddl/Members/members.riddl
```diff
@@ -5,6 +5,7 @@ type MemberMap is mapping from MemberId to MembersContext.Info
 
 context MembersContext  is {
     //Member Details
+    type StatusUpdates is any of {Active, Inactive, Suspended} briefly "Internal status of an Organization."
```
reid-spencer: This isn't well-named.  The confusing part for me is how an enumerator has anything to do with "Update". This is a representation of a state machine's state.  These aren't internal either, there ought to be public API to change and retrieve these states.  

How about "MemberState" ?

src/main/riddl/Members/members.riddl
```diff
@@ -146,30 +147,14 @@ context MembersContext  is {
...
-    event MemberInactivated is {memberId: MemberId, meta: MembersContext.MetaInfo}
-    briefly "Event issued after InactivateMember"
-    described as {
-      |This event is issued after InactivateMember to denote
-      |that the member is now inactive.
+    command UpdateMemberStatus is {
+      memberId: MemberId,
+      status: ^MemberContext.StatusUpdates,
+      updatingMember: MemberId
     }
```
reid-spencer, line 196: I actually prefer the previous arrangement. I would like to see a set of three commands, and corresponding events, for each of the "StatusUpdates" (MemberStates) enumerators.  I am going to object to any command that starts with "Update" or "Change".  They are are lazy verbs, a lot like "nice" and "good" are lazy adjectives.  They don't provide enough information.  For example, `InactivateMember` leads one to understand that a member is going to become inactive, just on the name alone, while `UpdateMemberStatus` doesn't give you as much information because you don't know what the member's status will be without looking at the RUNTIME value of a field of field that is not discernable at time of reading the code.  APIs need to be intuitive too.

src/main/riddl/Members/members.riddl
```diff
@@ -245,12 +230,6 @@ context MembersContext  is {
               and "set entity lastModifiedOn to today()"
               and "yield event MemberActivated"
             }
-            on command InactivateMember {
-              then error "Cannot set state to Inactive on a member that is not active"
-            }
-            on command SuspendMember {
-              then error "Cannot set state to Suspended on a member that is not active"
-            }
             on command TerminateMember {
```
reid-spencer: These need to be retained per the comment above about UpdateMemberStatus message

src/main/riddl/Members/members.riddl
```diff
@@ -277,19 +256,19 @@ context MembersContext  is {
...
             meta: MembersContext.MetaInfo
         }
-        state Active of ^ActiveState is {
-            invariant time_sanity is {
-              <=(@Active.meta.createdOn, @Active.meta.lastModifiedOn)
-            }
+        state ActiveMember of ActiveMemberState is {
+            //invariant time_sanity is {
+            //  <=(@ActiveMemberState.meta.createdOn, @ActiveMemberState.meta.lastModifiedOn)
+            //}
```
reid-spencer: You have removed this invariant. Please uncomment it and make it work

src/main/riddl/Members/members.riddl
```diff
@@ -313,21 +292,13 @@ context MembersContext  is {
...
-                on command SuspendMember {
-                  given "entity has suspensionReason non-empty"
-                  then "morph Member entity to Suspended state"
-                  and "set entity lastModifiedBy to actingMember from message"
-                  and "set entity lastModifiedOn to today()"
-                  and "yield event MemberSuspended"
+                on command UpdateMemberStatus {
+                    then "morph entity Member to state provided in UpdateMemberStatus"
+                    and "set Meta data to reflect the change"
+                    and "yield event MemberStatusUpdated"
```
reid-spencer: Again, I would prefer the individual handling of these three commands you have deleted.  If you want to keep it DRY, create a function in this entity to handle the common processing part.

src/main/riddl/Members/members.riddl
```diff
@@ -313,21 +292,13 @@ context MembersContext  is {
...
-                on command InactivateMember {
-                  then "morph Member entity to Inactive state"
-                }
-                on command SuspendMember {
-                  given "entity has suspensionReason non-empty"
-                  then "morph Member entity to Suspended state"
-                  and "set entity lastModifiedBy to actingMember from message"
-                  and "set entity lastModifiedOn to today()"
-                  and "yield event MemberSuspended"
+                on command UpdateMemberStatus {
```
reid-spencer: All three actions of the example's "then" part (below) should be specific operators, not arbitrary (string) ones. Please use the few operators we have when appropriate.  Furthermore, these three are not appropriate. You've opted for enumerator based state changes which means:
* You need to set the state value in a "MemterState" (the enumeration type) valued field
* Change the handler with the "become" operator.  Using "morph" means you're changing to a different RIDDL State

src/main/riddl/Members/members.riddl
```diff
@@ -313,21 +292,13 @@ context MembersContext  is {
...
+                    then "morph entity Member to state provided in UpdateMemberStatus"
+                    and "set Meta data to reflect the change"
+                    and "yield event MemberStatusUpdated"
                 }
                 on command TerminateMember {
-                  then "morph Member entity to Terminated state"
+                    then morph entity Member to state MemberTerminated with !TerminatedMemberState(lastMeta = @ActiveMemberState.meta)
```
reid-spencer: Good!

src/main/riddl/Members/members.riddl
```diff
@@ -338,103 +309,10 @@ context MembersContext  is {
...
-        } described as {
-          |This state describes that a member cannot access anything due
-          |to being a bad actor. This state transition is only possible
-          |as a command issued by an administrator and therefore can only
-          |be removed from the suspended state by an administrator.
-        }
-
-
-        record TerminatedState is {
-            memberId: MemberId // needed for compilation - RIDDL requires fields object and fields object requires at least one field
```
reid-spencer, line 435: The HANDLERS in the above-deleted states should probably be placed into the "ActiveMember" state you have defined above.  Copy and Paste probably won't work because you need to validate that the actions used in the Handler's Examples are correct for the new singular state they belong to.

reid-spencer: If there is no difference in behaviors between the various handlers then you don't need to move them. I just suspect there are some differences, especially in the manner in which certain commands are handled in the Inactive or Suspended states.

src/main/riddl/Organization/organization.riddl
```diff
@@ -20,16 +20,16 @@ context OrganizationContext is {
             on other { then error "Only EstablishOrganization command is allowed in the default state" }
         }
 
-        record DraftState is {
-            info: OrganizationContext.Info,
+        record DraftTenantState is {
+            info: OrganizationContext.OptionalInfo,
             parent: OrganizationId?,
```
reid-spencer: Shouldn't this also be in OptionalInfo since it is optional?

src/main/riddl/Organization/organization.riddl
```diff
@@ -20,16 +20,16 @@ context OrganizationContext is {
...
             members: MemberId+,
             owners: MemberId+,
             contacts: Contact?,
```
reid-spencer: Shouldn't this also be in OptionalInfo since it is optional?

src/main/riddl/Organization/organization.riddl
```diff
@@ -20,16 +20,16 @@ context OrganizationContext is {
             on other { then error "Only EstablishOrganization command is allowed in the default state" }
         }
 
-        record DraftState is {
-            info: OrganizationContext.Info,
+        record DraftTenantState is {
```
reid-spencer: Why is this named `DraftTenantState` ? Shouldn't it be `DraftOrganizationState` ?

src/main/riddl/Organization/organization.riddl
```diff
@@ -20,16 +20,16 @@ context OrganizationContext is {
...
+            info: OrganizationContext.OptionalInfo,
             parent: OrganizationId?,
             members: MemberId+,
             owners: MemberId+,
             contacts: Contact?,
             // accounts: OrganizationContext.AccountList,
             meta: OrganizationContext.MetaInfo
         }
-        state Draft of ^DraftState is {
+        state Draft of DraftTenantState is {
```
reid-spencer: Why is this named DraftTenantState ? Shouldn't it be DraftOrganizationState ?

src/main/riddl/Organization/organizationMessages.riddl
```diff
@@ -98,15 +94,35 @@ event OrganizationContactsUpdated is {
...
+command ActivateOrganization is {
+  orgId: OrganizationId,
+  updatingMember: MemberId
+}
+event OrganizationActivated is {
+  orgId: OrganizationId,
+  meta: OrganizationContext.MetaInfo
+}
+
 command UpdateOrganizationStatus is {
```
reid-spencer: I would prefer that this "Update..." form of command not occur, as I indicated in member.riddl.  Be specific in what it does so the reader "gets it" immediately. This kind of naming abstraction is not helpful downstream.  You already have "ActivateOrganization" above, follow that pattern for each kind of state change so the API is precise.

src/main/riddl/Organization/organizationTypeDefinitions.riddl
```diff
@@ -1,4 +1,4 @@
-type Status is any of {Draft, Active, Suspended, Terminated} briefly "Internal status of an Organization."
+type StatusUpdates is any of {Active, Suspended} briefly "Internal status of an Organization."
```
reid-spencer: Why not "Inactive" like you have for Members?

src/main/riddl/Tenant/tenantMessages.riddl
```diff
@@ -27,15 +33,24 @@ described by {
   |Given a TenantId, the tenant status is updated. This command yields a
   |TenantStatusUpdated event.
 }
-event StatusUpdated is {
+
+command TerminateTenant is {
+	tenantId: TenantId,
+	terminatingUser: String
+} briefly "A command that terminates a tenant bny deleting all data except its ID"
```
reid-spencer: bny → by

src/main/riddl/Tenant/tenantMessages.riddl
```diff
@@ -27,15 +33,24 @@ described by {
...
+described by {
+  |Given a TenantId, the tenant terminated. This command yields a
+  |TenantTerminated event.
```
reid-spencer: The description of a command should describe more than the `briefly` statement.  Perhaps you can explain why all data is deleted except the ID .. and is that even stored or just a field of the command?

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
```diff
@@ -1,3 +1,10 @@
+type OptionalInfo is {
+  name: String?,
+  primaryContact: Contact?,
+  address: Address?,
+  orgs: Organization*
```
reid-spencer, line 5: * != ?
I wouldn't put 0 or more in the category of "Optional"  I think "orgs" is a primary field, should be named organizations, and be placed into the Info of the Tenant.  The tenant's primary purpose is to "house" organizations and members.

reid-spencer: Another way to model memvbers and organizations here is as a query on a projection, provided as a function, because each member and organization references its owning tenant so the projection should be able to get them for use in tenant by something like "SELECT id from members" and similarly for organizations.

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
```diff
@@ -30,8 +37,7 @@ described as {
   |in the Info type."
 }
 
-type Status is any of {
-  Draft briefly "Tenant is inactive, in process of being created/defined",
+type StatusUpdates is any of {
```
reid-spencer: Similarly as for members, I don't think this name fits.  How about TenantState?

reid-spencer (changes requested):
There are several things that need correcting or considering. 

Also, in this RIDDL based repo, I would prefer that PRs contain themselves to one bounded context.  This one considers 3 and it gets hard for the reviewer to maintain context well.

src/main/riddl/Members/members.riddl
AlexWeinstein92: In all places I've touched there is a command to change state. In the other 2 BCs (Org & Tenant) the status is included in MetaInfo and can be retrieved from there.

I can change all Status types according to `[Entity]State` pattern, add Status to MetaInfo of Member if desired, and add explicit commands for retrieving states if desired.

The only quibble I have with those is that I want to keep the word `Status` here to differentiate Active enum from `record Active[Entity]State` that may contain one of these enums. Unless we want to eliminate the name `Active` from those `State` records... I've considered just calling them `[Entity]State` to eliminate possible confusion of having 2 `Active` keywords in the same place.

I also want to keep the `Status` keyword because these are not the same as the RIDDL record-based states.

src/main/riddl/Members/members.riddl
AlexWeinstein92: I can change back but don't really understand the reasoning here. `UpdateMemberStatus` will always contain what the member's status will be in the `status` field. A command with a field that says the Status type doesn't seem much worse than a command named with that Status type. The runtime value will result in a success or error regardless of this choice. It would be probably easier to pair results/errors with commands though.

src/main/riddl/Organization/organization.riddl
AlexWeinstein92: Both this and the other one you noted are desired by James to be on top level. This was established in conversations a while ago. We don't want them in Info for various reasons, mostly to do with importance of use and so inappropriateness of putting them in Info.

In Draft state they are Optional so that we can check if they are there when `Activate` command is sent

src/main/riddl/Members/members.riddl
AlexWeinstein92: This was a big oversight - will look into it

src/main/riddl/Organization/organizationTypeDefinitions.riddl
AlexWeinstein92: Inactive is different from Suspended for Members in that Suspended is an explicit action to be taken by an admin, while Inactive is a state the system uses when a Member has not been using the app for a while. Not sure if we want to measure the same for Orgs, but I don't see why not really.

src/main/riddl/Tenant/tenantMessages.riddl
AlexWeinstein92: It's not stored in state record because it is already present given that we are still in an entity handler

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
AlexWeinstein92: Orgs is already in Info of Tenant, this is just the version of Info with all fields Optional. I agree a projection may be more appropriate, but at the same time it may make sense to have Organizations in Tenant Info just in case there are scenarios when we want to get Orgs directly from Tenant without having to go to a View (I'm not sure why though...)

src/main/riddl/Members/members.riddl
reid-spencer: re:
> I can change all Status types according to [Entity]State pattern, add Status to MetaInfo of Member if desired, and add explicit commands for retrieving states if desired.

Sounds about right. Let's not bother adding the explicit commands for now until there is a demonstrated reason to do so. We don't have a use case for it so lets wait until there is one. 

> The only quibble I have with those is that I want to keep the word Status here to differentiate Active enum from record Active[Entity]State that may contain one of these enums ...

So, there is quite a conundrum here because of the dual nature of RIDDL's state machine support.  I would suggest you think about RIDDL's "state" keyword at the meta-level. In the application that comes out from it, that meta-level disappears.  The "Status" you refer to are the definitional states of a state machine and so I prefer the word "state" for this. If you look at the definitions of these words, state can be attributed to inanimate or animate things. But status is mostly about people, law, society or a situation. The definition of "State" include:
* A condition of being in a stage or form, as of structure, growth, or development.
* The condition of a physical system with regard to phase, form, composition, or structure.
That fits a state machine pretty well.  "Status" on the other hand doesn't have an equivalent. All of its connotations are these:
* Position relative to that of others; standing.
* High standing; prestige.
* The legal character or condition of a person or thing.
* The state of affairs; the situation: synonym: state.
* Standing or position as regards rank or condition.
* Position of affairs.
* In law, the standing of a person before the law in the class of persons indicated by his or her legal qualities; the relation fixed by law in which a person stands toward others or the state.
* A person’s [condition](https://www.wordnik.com/words/condition), [position](https://www.wordnik.com/words/position) or [standing](https://www.wordnik.com/words/standing) relative to that of others.
The thing that I am concerned about with Status is its relative nature as in "standing", "position relative to", "relative to that of others".  State doesn't impact any of this relativeness to me. 

Consider the above and then decide for yourself, we don't need to "quibble" further.

src/main/riddl/Members/members.riddl
reid-spencer: There is no way to determine the status implied by the UpdateMemberStatus definition. All you know is that the status will change (perhaps), but you don't know what it will become. Inspection of its definition of that field still doesn't tell you the ACTUAL value that will be set, only that it is of "Status" type. ActivateMember is very clear. You don't even have to look at the content of the message.  What if, later on, you wanted to change the representation of the status from an enum to a set of records (not RIDDL states). You would have to change UpdateMemberStatus to expose internal details of the entity. That change could have a huge ripple effect on all the users of Member.  With ActivateMember you don't have to do this and it has no ripple effect. It is just an encapsulated implementation detail.  APIs need to be designed defensively, just like driving a car.  If you still don't get it, please just trust me. This is the right way to design an API.

src/main/riddl/Members/members.riddl
AlexWeinstein92: The keyword may disappear, but the records themselves have State in the name. However I do prefer dropping the dual application of `Active` in naming conventions. I just went through some possible word choices and I think if we call them `Established[Entity]State` instead of `Active[Entity]State` it will work well because we usually use `Establish[Entity]` command to change to enumerated state. In Event I can use `ScheduledEventState`, and things like that where appropriate. 

I will change the `UpdateStatuses` to `[Entity]State` as well.

src/main/riddl/Organization/organizationTypeDefinitions.riddl
reid-spencer: It seems that "Inactive", per your definition,  could be a state that is modeled in Akka as "passivated" but I wouldn't want to make a state change just to passivate an actor. To disambiguate, we need to understand the behavior differences are in the "Inactive" state. If Member doesn't behave any differently between Active or Inactive then Inactive isn't useful.  If that's the case, then remove Inacive from Members and don't add it to any entity definition. The only reason to have a State Machine state is to capture either data or behavior differences.  If neither changes, the state isn't needed.

src/main/riddl/Tenant/tenantMessages.riddl
reid-spencer: Right. So please fix the doc in `briefly`

src/main/riddl/Members/members.riddl
AlexWeinstein92: Ok, that makes total sense, being cautious and thinking ahead :)

src/main/riddl/Organization/organizationTypeDefinitions.riddl
AlexWeinstein92: It's important to keep in mind that there is a difference between a state that uses its own handlers and a state that is enumerated and so shares handlers with other states. Hence the desire to disambiguate using `Status`es for the latter.

Inactive and Suspended and both cases of the latter, so we will not be changing states in a heavy way, only by noticing that the enum has a certain value and thus hitting a point in the handler flow that passivates the state.

src/main/riddl/Tenant/tenantMessages.riddl
AlexWeinstein92: Already done :) just not pushed yet

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
reid-spencer: Okay, I couldn't see the Info record from the brief snippet provided.  Since this is the OptionalInfo, I wouldn't include "orgs" in it.  That's because having organizations is not a prerequisite of a complete "Tenant".  The tenant can be active with no organizations. It may be just a "members only club" :)  This rule is true for all of these "OptionalInfo" records for "Draft" state in any entity.  That is: "Draft" state only needs to model the required fields of the permanent state.  `name` is required to generate a log message or error message and indicate where it came from.  `address` is required by billing because the tenant will also be the means of payment.  `primaryContact` is the person who is responsible for payment, and thus required.   But organizations aren't required for a Tenant to be a Tenant.

AlexWeinstein92: I was thinking the required fields could easily be discerned, but not directly from the OptionalInfo, rather by comparing to Info. Since Organizations are `*` and not `+` in Info, it is not a required field. That may not be obvious enough though?

src/main/riddl/Members/members.riddl
reid-spencer: Thanks, and I agree with avoiding dual use of Active.  Established is fine if "Active" is one of enumerated states.

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
reid-spencer: Yeah, totally not obvious enough. But more importantly, it blurs the state transition. How many organizations are needed before the entity can transition from Draft to Established?  1? 10? 100?  Its best not to deal with these kinds of issues and recognize in Draft state that organizations are not essential.  If this entity wanted to auto-transition to established, it would have to have some minimum number of "orgs" needed.  With merely optional fields, there are only two cardinalities: 0 (not complete) and 1 (complete). So, when all optional fields are "complete", the entity can auto-transition to Established.

src/main/riddl/Organization/organizationTypeDefinitions.riddl
reid-spencer: Agreed, but did you get my point above?

I don't think "Inactive" is a valid state. Unless you can show a case where "Inactive" has a handler that is definitionally different than that of the other states, it should just be removed, everywhere.  I'm basing this on your assertion that "Inactive is a state the system uses when a Member has not been using the app for a while".  In other words, this is time-out transition the system takes in that entity on its own. There is no command needed externally to force the transition to Inactive.  This happens during passivation of actors.  We don't need this state in our model because it has nothing to do with the business logic .... UNLESS you can show how the Inactive entity responds to messages differently than Active ones.  If the Active and Inactive states are identical then one of them is redundant and I'm suggesting removal of "Inactive".

AlexWeinstein92: That does make sense. There is no different business logic, the difference in state is entirely technical. I hadn't thought about it that way.

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
AlexWeinstein92: So are you recommending that for `DraftState`s we have 2 `DraftInfo` objects, one for Required fields and the other for Optional? And then we combine when creating Info Established or other subsequent state? This would of course require bringing back the `UpdateInfo` object name in place of what we have as `OptionalInfo` now (with the goal of using the options to denote whether an Info field has been filled in as part of an update request)

src/main/riddl/Organization/organization.riddl
reid-spencer: I wasn't suggesting the "contacts" field go in Info, but in OptionalInfo, since it is optional.  Looking at this further, the Draft's "info" field is OptionalInfo, which seems weirdly named.  Given that there are 3 other fields in play here, why bother with any OptionalInfo field?

AlexWeinstein92: I think this will be resolved once you confirm the question here https://github.com/improving-app/riddl/pull/45#discussion_r1145278433

src/main/riddl/Tenant/tenantTypeDefinitions.riddl
reid-spencer: I can't say categorically. I look at it this way:
* Draft is the first state defined, so therefore it is the state that occurs immediately after creation. 
* Initialization of the entity can be handled in two ways: immediately or lazily
* If immediately, then the EstablishFoo command handled by the initialization handler (at the entity level) is responsible for completely initializing Foo's state, presumably from EstablishFoo's fields. In this case naming the first state "Draft" is misleading. 
* If lazily, then the EstablishFoo command might initialize Foo's Draft state but it doesn't become usable until it transitions to Established (as we discussed) and a handler that does more than initialize Draft becomes active.

AlexWeinstein92: Right, I was only considering the lazy case here :)

src/main/riddl/Members/members.riddl
```diff
@@ -37,7 +65,7 @@ context MembersContext  is {
       |of the Member with demographics, contact info, and various options
       |like notifications and group membership the Member can opt in to.
     }
-    type OptionalInfo is {
+    type EditInfo is {
```
reid-spencer: The name occurs for me as a bit weird. Maybe "EditableInfo" because it is the info you can address with the EditMemberInfo command?

src/main/riddl/Members/members.riddl
```diff
@@ -237,7 +270,7 @@ context MembersContext  is {
               and "set entity lastModifiedOn to today()"
```
reid-spencer: If you want to make this a real set, now() and today() are both recognized functions that return TimeStamp and Date respectively.

src/main/riddl/Members/members.riddl
```diff
@@ -256,19 +289,19 @@ context MembersContext  is {
...
             info: MembersContext.Info,
             meta: MembersContext.MetaInfo
         }
-        state ActiveMember of ActiveMemberState is {
-            //invariant time_sanity is {
-            //  <=(@ActiveMemberState.meta.createdOn, @ActiveMemberState.meta.lastModifiedOn)
-            //}
+        state RegisteredMember of RegisteredMemberState is {
+            invariant time_sanity is {
+              <=(@RegisteredMemberState.meta, @RegisteredMemberState.meta)
```
reid-spencer, line 398: Let's remember that this needs to be modified after I've fixed the validation bug.

src/main/riddl/Organization/organization.riddl
```diff
@@ -12,32 +12,97 @@ context OrganizationContext is {
...
+                    ),
+                    meta = !OrganizationContext.MetaInfo(
+                        createdOn = now(),
+                        createdBy = @EstablishOrganization.establishingMember,
+                        lastUpdatedOn = now(),
+                        lastUpdatedBy = @EstablishOrganization.establishingMember,
+                        currentState = @OrganizationStates.Draft
+                    )
+                )
             }
```
reid-spencer, line 54: Nice!

src/main/riddl/Organization/organization.riddl
```diff
@@ -341,26 +406,46 @@ context OrganizationContext is {
...
+            handler SuspendedOrganizationHandler is {
+                on command EstablishOrganization {
+                    then error "Organization already established"
+                }
+                on command ActivateOrganization {
+                    then become entity Organization to handler ActiveOrganizationHandler
+                    and "update the meta info to reflect the change"
+                    and "yield event OrganizationActivated"
+                }
```
reid-spencer, line 500: There probably needs to be some other kind of check for this command. Something that answers: "is this entity allowed to activate again?" .. just a though, not sure what should be checked but it seems likely. It was suspended fora  reason. Has that reason disappeared?

src/main/riddl/Organization/organization.riddl
```diff
@@ -341,26 +406,46 @@ context OrganizationContext is {
...
+                 on command SuspendOrganization {
+                    given "entity has suspensionReason non-empty"
+                    then "morph Organization entity to Active state"
```
reid-spencer: `to Active state` ? This command is to suspend the organization while it is already in suspended state.  So, shouldn't it no-op? or at least just update the suspension reason?

src/main/riddl/Organization/organizationSources.riddl
```diff
@@ -1,7 +1,9 @@
...
     OrganizationEstablished,
-    OrganizationInfoUpdated,
-    OrganizationStatusUpdated,
+    OrganizationActivated,
+    OrganizationSuspended,
+    OrganizationTerminated,
+    OrganizationInfoEdited,
     MembersAddedToOrganization,
     OwnersAddedToOrganization,
     OrganizationContactsUpdated,
```
reid-spencer, line 9: Nice to see an alternation type being used!

src/main/riddl/Organization/organizationTypeDefinitions.riddl
```diff
@@ -22,8 +39,8 @@ type Info is {
     //color scheme
     //calendar: Calendar
 } briefly "Essential information about an organization that is expected to evolve over time."
-// TODO: We added this UpdateInfo definition for the UpdateOrganizationInfo command. This definition makes everything optional.
-type OptionalInfo is {
+
+type EditInfo is {
```
reid-spencer: EditableInfo ?

src/main/riddl/Tenant/tenantProjections.riddl
```diff
@@ -17,11 +17,11 @@ projector TenantViews is {
...
-        on event TenantContext.InfoUpdated {
-            then "find tenant set to TenantEstablished.tenantId"
-            and "if name has changed, set tenantName to TenantEstablished.info.name"
-            and "set tenantInfo to TenantEstablished.info"
-        }
+        //on event TenantContext.InfoUpdated {
+        //    then "find tenant set to TenantEstablished.tenantId"
+        //    and "if name has changed, set tenantName to TenantEstablished.info.name"
+        //    and "set tenantInfo to TenantEstablished.info"
+        //}
```
reid-spencer: Why is this commented out?

reid-spencer (changes requested):
Mostly LGTM. A few naming choices to ponder, a couple questions below.

src/main/riddl/Members/members.riddl
AlexWeinstein92: Once it's validating types properly, all these lines where I don't go into the 3rd field I need to should error out, which will tell me what I need to change. If I am mistaken about that, I think I have a pretty good idea anyway of what I've changed and so where I need to look.

src/main/riddl/Organization/organization.riddl
AlexWeinstein92: The `reason` is just a `String`, so I think unless we get an NLP model in there or find a way to use an enum (which I don't see likely to be easy at all), the reasons will just be there for the logs. I can check with James though and see what his intentions were with this.

src/main/riddl/Organization/organization.riddl
AlexWeinstein92: Likely a copy paste error, should just update the reason. Note that this is code I brought back from before I modified anything - this is what it looks like on main. Will fix it.

src/main/riddl/Tenant/tenantProjections.riddl
AlexWeinstein92: InfoUpdated was only used on Draft state. Really I should have asked if we want to be able to edit Tenant Info after creation, once we decided not to have Tenant Draft state. Too much going on, got left behind.

src/main/riddl/Organization/organization.riddl
reid-spencer: okay, good enough for now. please ignore

src/main/riddl/Tenant/tenantProjections.riddl
reid-spencer: Yes, Tenant Info needs to be editable once Established.  For example, the primary contact person changes or the credit card changes or ... . 

So, yes, let's not comment it out because TenantInfoUpdated event needs to still be handled

reid-spencer (changes requested):
Just need one commented out thing uncomments in tenantProjections. 

THen you can merge this.

src/main/riddl/Tenant/tenantProjections.riddl
AlexWeinstein92: This is still the command for Draft state. But I will add it to Established SuccessHandler

src/main/riddl/Tenant/tenant.riddl
```diff
@@ -28,129 +46,158 @@ context TenantContext is {
...
+        //        //         then error "DistributionAccount ID does not exist"
+        //        //     }
+        //        //     example NotAuthorized {
+        //        //         when "updating user is not authorized to make changes to the tenant"
+        //        //         then error "User is not authorized to modify Tenant"
+        //        //     }
+        //        // }
+        //        on other { then error "Command not allowed in Draft state" }
+        //    }
+        //}
```
reid-spencer, line 163: Because the above is double commented out, you're effectively deleting this state?

reid-spencer (approved):
LGTM .. just one question about a large amount of code commented out.

src/main/riddl/Tenant/tenant.riddl
reid-spencer: @AlexWeinstein92 - this question is blocking the merge

AlexWeinstein92: Yes, this state is non-existent after being commented out. The double comments are because there was previously commented out code in the state.

I am only not deleting it in case James decides for some reason to have Draft for Tenant.

reid-spencer: Okay.

PR #57, opened 2023-05-31
fix: required fields & update to riddlc 0.23 [IA-171]

AlexWeinstein92:


src/main/riddl/Common/Contact.riddl
```diff
@@ -1,13 +1,13 @@
 // Copyright (c) 2022 Improving Inc.
 
 // Types for Contact Information
-type MobileNumber is Pattern("\\(([0-9]{3})\\)([0-9]{3})-([0-9]{4})")
+type MobileNumber is Pattern("(([0-9]{3}))([0-9]{3})-([0-9]{4})")
```
reid-spencer: This is incorrect. 
a ( is a special character (grouping0 in a regular expression. To get a literal character, as in the original, you need to prefix it with a \ but that is a special character inside "" so you need another \ to escape the backslash and end up with regex seeing \( as a literal ( character. This is well documented in the Java Regex documentation

src/main/riddl/Common/Contact.riddl
```diff
@@ -1,13 +1,13 @@
...
+type EmailAddress is Pattern("(?:[a-z0-9!#$%&'*+/=?^_`{|}~-](?:.[a-z0-9!#$%&'*+/=?^_`{|}~-])|\"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])\")@(?:(?:[a-z0-9](?:[a-z0-9-][a-z0-9])?.)+[a-z0-9](?:[a-z0-9-][a-z0-9])?|[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?).){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)])")
   briefly "RFC 5322 official standard. See https://emailregex.com"
-type CAPostalCode is Pattern("[A-Z]\\d[A-Z]\\d[A-Z]\\d")
+type CAPostalCode is Pattern("[A-Z]d[A-Z]d[A-Z]d")
   briefly "A Canadian Postal Code'"
-type USPostalCode is Pattern("\\d{5}(-\\d{4})?")
+type USPostalCode is Pattern("d{5}(-d{4})?")
   briefly "A USA Postal Code"
```
reid-spencer: The above changes are also problematic for the same reasons.

reid-spencer (changes requested):
The regex changes you made are completely wrong and should all be reverted. 

The other changes to the various actions I don't understand and don't seem correct to me. Please write a description for this PR so I have some understanding of what you're trying to accomplish. Also, please note thenew features in0.23.0: you can name the message referenced in an On Clause and use it in expressions in the examples defined in that clause.

src/main/riddl/Common/Contact.riddl
AlexWeinstein92: It must have been something wrong with version 0.22 because that was the only way I could get it to compile. Now it seems to be working the proper way. Will push the fix

src/main/riddl/Organization/organizationMessages.riddl
```diff
@@ -145,3 +145,39 @@ result OrganizationInfo is {
   orgId: OrganizationId,
   info: OrganizationContext.Info
 }
+query GetOrganizationParent is {orgId: OrganizationId}
+result OrganizationParent is {
+  orgId: OrganizationId,
+  parent: OrganizationId?
```
JamesYopp: Why is parent optional? If parent doesn't exist shouldn't the result be an error?

src/main/riddl/Organization/organizationMessages.riddl
```diff
@@ -145,3 +145,39 @@ result OrganizationInfo is {
...
+}
+query GetOrganizationContacts is {orgId: OrganizationId}
+result OrganizationContacts is {
+  orgId: OrganizationId,
+  contacts: Contact*
+}
+query GetOrganizationMeta is {orgId: OrganizationId}
+result OrganizationMeta is {
+  orgId: OrganizationId,
+  meta: OrganizationContext.MetaInfo?
```
JamesYopp: I don't think Meta can be optional. How does it not exist?

AlexWeinstein92: I was doing that without thinking. Good catch - thought I got them all

src/main/riddl/Organization/organizationMessages.riddl
```diff
@@ -145,3 +145,39 @@ result OrganizationInfo is {
...
+}
+query GetOrganizationState is {orgId: OrganizationId}
```
JamesYopp, line 166: This was confusing to me. When I first read this I was thinking Open, Closed, etc. But I can see by the result that you are thinking all the attributes that define that organization. Wouldn't those attributes potentially change by state (as in Open, closed)?

src/main/riddl/Store/Store.riddl
```diff
@@ -17,35 +17,24 @@ context StoreContext is {
...
-
-  type OptionalDraftInfo is {
-    event: EventId?
-  }
-
   type Info is {
-    name: String,
-    description: String,
-    products: SKU+,
+    name: String briefly "Required",
```
JamesYopp, line 21: Maybe we need to think about "required". By declaring "name: String" it is by definition required (as opposed to "name: String?" or "name: String*". Perhaps there is a conversation at the RIDDL level? What is the benefit of adding "briefly "Required""?

src/main/riddl/Organization/organizationMessages.riddl
AlexWeinstein92: If parent doesn't exist we are at top level (I assume since every state already has it as optional -may also be used to handle draft state option)

src/main/riddl/Organization/organizationMessages.riddl
AlexWeinstein92: Hence optionality or `*` instead of `+` on those fields

src/main/riddl/Store/Store.riddl
```diff
@@ -348,18 +336,18 @@ context StoreContext is {
...
+              location = @CreatedStoreState.info.location,
+              sponsoringOrg = @CreatedStoreState.info.sponsoringOrg,
               meta = @CreatedStoreState.meta
             )
           }
           on command OpenStore {
             then set CreatedStoreState.meta to !StoreContext.MetaInfo(
-              createdOn = @CreatedStoreState.meta,
-              createdBy = @CreatedStoreState.meta,
+              createdOn = @CreatedStoreState.meta.ceratedOn,
```
JamesYopp: typo - createdOn not ceratedOn

src/main/riddl/Organization/organization.riddl
```diff
@@ -236,25 +228,26 @@ context OrganizationContext is {
...
-                               url = @DraftOrganizationState.optionalInfo,
-                               logo = @DraftOrganizationState.optionalInfo,
-                               tenant = @DraftOrganizationState.requiredInfo
+                               name = @DraftOrganizationState.info.name,
+                               shortName = @DraftOrganizationState.info.shortName,
+                               address = @DraftOrganizationState.info.address,
+                               isPrivate = @DraftOrganizationState.info.isPrivate,
+                               url = @DraftOrganizationState.info.url,
+                               logo = @DraftOrganizationState.info.logo,
+                               tenant = @DraftOrganizationState.inf0.tenant
```
JamesYopp: typo - info vs inf0

src/main/riddl/Organization/organizationMessages.riddl
JamesYopp: Thinking out loud... What if the top level assumed itself as parent? 

This is in a result definition so I suppose it doesn't matter much. Just feels weird to have parent be optional.

src/main/riddl/Organization/organizationMessages.riddl
JamesYopp: I get your point. But just to be pedantic: + would be required (one or more)
I think optionality would be * (zero or more) or ? (zero or one).

I am not sure if this distinction would be confusing to others or not. The way I was thinking in the first iteration:
getState = returns active, inactive, suspended, terminated, etc.
getInfo = returns data about the organization (info, metaInfo, etc.)

src/main/riddl/Store/Store.riddl
AlexWeinstein92: The comment is an instruction for the developer that this field has to be present in a `DraftState`'s `EditableInfo` before moving out of that state, and cannot be removed if out of that state (though it can still be edited)

reid-spencer: The `briefly` description is redundant with `name: String`, which is an instruction to the developer that this field is required and cannot be omitted.

AlexWeinstein92: I agree that it is self-evident, but developers were confused at one point by lack of special character for required fields. This is why I created `RequiredDraftInfo` and `RequiredOptionalInfo`, which are being removed here. These annotations help point out to developers that these are Required fields when they are new to RIDDL (and possibly protobuffers & scala as well)

src/main/riddl/Organization/organizationMessages.riddl
AlexWeinstein92: Have changed the result to only return `currentState` contained in `metaInfo`

src/main/riddl/Organization/organizationMessages.riddl
AlexWeinstein92: Per our conversation weeks ago, I am changing this to return parent without option. I am not sure though where to document that parent can be same as self orgid

JamesYopp (approved):
Thanks Alex.

reid-spencer (approved):
LGTM

PR #60, opened 2023-06-23
feat: product updates & some Store cleanup [IA-203]

AlexWeinstein92:
- Refactoring of Product service to properly make use of state `record`s and `handler`s
- Deleted->Terminated in Product
- completing paths in Store service


JamesYopp (approved):
Thanks Alex

PR #61, opened 2023-07-05
Complete RIDDL for Event (IA-204)

AlexWeinstein92:
Entity records & their handlers need to be refactored to accommodate state changes appropriately

### DIFF
--- a/src/main/riddl/Events/eventContext.riddl
+++ b/src/main/riddl/Events/eventContext.riddl
@@ -253,10 +253,9 @@ context EventContext is {
                  expectedStart = @DraftEventState.info.expectedStart,
                  expectedEnd = @DraftEventState.info.expectedEnd,
                  isPrivate = @DraftEventState.info.isPrivate,
-                 eventStatusInfo = !CancelledEventInfo(
-                   reason = @CancelEvent.reason,
-                   timeStarted = startTimeIfPast(startTime = @ScheduledEventState.info.expectedStart),
-                   timeEnded = nowIfStartIsPast(startTime = @ScheduledEventState.info.expectedStart)
+                 eventStatusInfo = !DelayedEventInfo(
+                   reason = @DelayEvent.reason,
+                   timeStartedOpt = startTimeIfPast(startTime = @ScheduledEventState.info.expectedStart)
                  )
             ),
             meta = !EventMetaInfo(
@@ -285,6 +284,11 @@ context EventContext is {
       //reservation: ReservationId?
     }
     state ScheduledEvent of Event.ScheduledEventState is {
+        handler DefaultScheduledEventHandler {
+             on other {
+                then "pick the correct handler based on meta.currentSate and become it"
+             }
+        }
         handler ScheduledEventHandler {
             on command ScheduleEvent{then error "Event is already scheduled. Please RescheduleEvent to change event information."}
             //on command AddReservationToEvent {
@@ -460,10 +464,9 @@ context EventContext is {
                         duration = @DelayEvent.expectedDuration
                       ),
                       isPrivate = @ScheduledEventState.info.isPrivate,
-                      eventStatusInfo = !CancelledEventInfo(
-                        reason = @CancelEvent.reason,
-                        timeStarted = startTimeIfPast(startTime = @ScheduledEventState.info.expectedStart),
-                        timeEnded = nowIfStartIsPast(startTime = @ScheduledEventState.info.expectedStart)
+                      eventStatusInfo = !DelayedEventInfo(
+                        reason = @DelayEvent.reason,
+                        timeStartedOpt = startTimeIfPast(startTime = @ScheduledEventState.info.expectedStart)
                       )
                     ),
                     meta = !EventMetaInfo(
--- a/src/main/riddl/Events/eventContext.riddl
+++ b/src/main/riddl/Events/eventContext.riddl
@@ -2,195 +2,666 @@ type EventId is Id (ImprovingApp.Events.EventContext.Event)
   briefly "Unique identifier for Tenant Entity"
 
 context EventContext is {
-    include "eventMessages"
-    include "eventTypeDefinitions"
-    include "eventSources"
+  include "eventMessages"
+  include "eventTypeDefinitions"
+  include "eventSources"
 
-    entity Event is {
-        option is aggregate
+  function incrementDelayedEventTime is {
+    requires {originalTime: DateTime, duration: Duration}
+    returns { newTime: DateTime }
+  } briefly "used for incrementing start time and end time when DelayEvent is received"
 
-        handler EventHandler is {
-            on command ScheduleEvent {
-                then morph entity Event to state Scheduled with
-                  !Event.ScheduledState(
-                    eventInfo = !EventInfo(),
-                    eventMetaInfo = !EventMetaInfo(),
-                    reservation = !ReservationId()
-                  )
-            }
-            on other {
-            then error "You must first create an event using ScheduleEvent command."
-            }
-        }
+  function startTimeIfPast is {
+    requires {startTime: DateTime}
+    returns { timeOpt: DateTime? }
+  } briefly "returns startTime if it is past, empty option if it isn't. This is used when CancelEvent is received"
 
-        record ScheduledState is {
-          eventInfo: EventInfo,
-          eventMetaInfo: EventMetaInfo,
-          reservation: ReservationId?
-        }
-        state Scheduled of Event.ScheduledState is {
-            handler ScheduledEventHandler {
-                on command ScheduleEvent{then error "Event is already scheduled. Please RescheduleEvent to change event information."}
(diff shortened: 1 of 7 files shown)


src/main/riddl/Events/eventContext.riddl
```diff
@@ -2,195 +2,666 @@ type EventId is Id (ImprovingApp.Events.EventContext.Event)
...
-    include "eventTypeDefinitions"
-    include "eventSources"
+  include "eventMessages"
+  include "eventTypeDefinitions"
+  include "eventSources"
 
-    entity Event is {
-        option is aggregate
+  function incrementDelayedEventTime is {
+    requires {originalTime: DateTime, duration: Duration}
```
JamesYopp, line 10: If I am thinking of this as a function call, I would expect that you send in the EventId and a duration. The EventId would have the original time. We may want to think about duration for a minute. Is it delayed in minutes, hours, days, weeks?

If we are sending this as a message to an actor, the actor would be an instance of a specific event so no event reference would be necessary. Again OriginalTime would not be necessary because the Event would have that info.

If we are looking at a function on a time calculator. Given this DateTime, and a delay of time, return what would be the new time, then I think this makes sense.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -2,195 +2,666 @@ type EventId is Id (ImprovingApp.Events.EventContext.Event)
...
+  function newEditableInfoFromFieldsPresentInEditable is {
+    requires { editable: EditableEventInfo, oldInfo: EditableEventInfo }
+    returns { newInfo: EditableEventInfo }
+  } briefly "returns newInfo with all fields present in editable and all those not present there taken from oldInfo"
+
+  function newInfoFromFieldsPresentInEditable is {
+    requires { editable: EditableEventInfo, oldInfo: EventInfo }
+    returns { newInfo: EventInfo }
+  } briefly "returns newInfo with all fields present in editable and all those not present there taken from oldInfo"
+
```
JamesYopp, line 33: I am not sure of the purpose of the above functions. I am sure it will make more sense when I get to the context of where they are used. But just reading through the changes I am not sure why we need these.

src/main/riddl/Events/eventContext.riddl
```diff
@@ -2,195 +2,666 @@ type EventId is Id (ImprovingApp.Events.EventContext.Event)
...
+  entity Event is {
+    option is aggregate
+
+    handler EventHandler is {
+      on command CreateEvent {
+        then send event EventCreated(
+            eventId = @CreateEvent.eventId,
+            info = @CreateEvent.info,
+            meta = !EventMetaInfo(
```
JamesYopp, line 42: Looks like new RIDDL syntax that I need to get caught up on. What does !EventMetaInfo do? And how is that different than @CreateEvent.info (other than the specific reference details)?

src/main/riddl/Events/eventTypeDefinitions.riddl
```diff
@@ -2,12 +2,34 @@ type LiveUpdate is mapping from String to String//i.e. "Score" -> "12 to 10", "a
 type EventResults is mapping from String to String //i.e. "Score" -> "12 to 10", "attendance" -> "523", ...
 
 /*
-type EventSpecificDetails is one of {FootballGame, BasketballGame, Play, Concert, FundRaiser}
+type Score is {home: Integer, away: Integer}
```
JamesYopp, line 5: I think we could comment this out for now

src/main/riddl/Events/eventTypeDefinitions.riddl
```diff
@@ -2,12 +2,34 @@ type LiveUpdate is mapping from String to String//i.e. "Score" -> "12 to 10", "a
...
+type Teams is {home: Team, away: Team}
```
JamesYopp, line 6: I think we could comment this out for now

src/main/riddl/Events/eventTypeDefinitions.riddl
```diff
@@ -2,12 +2,34 @@ type LiveUpdate is mapping from String to String//i.e. "Score" -> "12 to 10", "a
...
+    teams: Teams,
+    ???
 }
+type FootballGameStateInfo is one of {ScheduledFootballGameInfo | InProgressFootballGameInfo | DelayedFootballGameInfo | PastFootballGameInfo | CancelledFootballGameInfo}
+type ScheduledFootballGameInfo is {}
+type InProgressFootballGameInfo is {yardsSoFarHome: Integer, yardsSoFarAway: Integer}
+type DelayedFootballGameInfo is {yardsSoFarHome: Integer?, yardsSoFarAway: Integer?}
+type PastFootballGameInfo is {totalYardsHome: Integer, totalYardsAway: Integer}
+type CancelledFootballGameInfo is {yardsSoFarHome: Integer?, yardsSoFarAway: Integer?}
+
```
JamesYopp, line 32: I think we could comment out all the FootballGame type details

src/main/riddl/Members/members.riddl
JamesYopp, line 1: I probably would have put these changes in a separate PR. So, let's have a PR for all Event related details and another PR for Member details, and so on. That is probably a matter of style more than anything. As I was doing the review it took a few seconds for me to switch my brain from thinking about Events to thinking about Members.

src/main/riddl/Product/product.riddl
JamesYopp, line 1: I probably would have put these changes in a separate PR. So, let's have a PR for all Event related details and another PR for Product details, and so on. That is probably a matter of style more than anything. As I was doing the review it took a few seconds for me to switch my brain from thinking about Events to thinking about Products. In this case the change is so small that it is questionable that a separate PR is useful. Then again, maybe not. Kind of thinking out loud.

src/main/riddl/Events/eventContext.riddl
AlexWeinstein92: This is not an external request of an kind, just a utility function for increasing a time value by a duration when you have it. If you look at the example, I am only using it as a transforming function on the time the Event was original at. In other words, this is called form inside an Event entity to generate the delayed time value.

src/main/riddl/Events/eventContext.riddl
AlexWeinstein92: `@` references an object in-memory (usually a command or state)
`!` creates a new object from the definition provided, and returns it as a value to the field or function calling it.

I am not sure why `EventCreated` does not need an exclamation mark here. Would have to ask Reid about that.

src/main/riddl/Events/eventTypeDefinitions.riddl
AlexWeinstein92: `/*` and `*/` indicate the beginning and end of comment blocks - in other words, already done :)

src/main/riddl/Product/product.riddl
AlexWeinstein92: I will revert the changes in both files after dropping them in new branches and pushing as small PRs

src/main/riddl/Events/eventContext.riddl
JamesYopp: Are you just using this inside of the DelayEvent message handler? If it is used in just one place I question the value of a function. I guess if you have multiple states that have handlers that must process the DelayEvent command then this makes better sense. 

What of the concern about the measure of time? I think what I am missing here is the definition of Duration.

src/main/riddl/Events/eventContext.riddl
JamesYopp: Good to know. Thanks!

src/main/riddl/Events/eventContext.riddl
AlexWeinstein92: Yes, it is used in multiple state handlers. Duration is, according to the riddl docs `the amount of time between a start and stop time`. This is a basic type in RIDDL as well as in most programming languages - there, it is usually measured in milliseconds, but nanoseconds are also possible.

src/main/riddl/Events/eventTypeDefinitions.riddl
JamesYopp: hah! I totally missed that. I was paying attention to the change in the line (removal of EventSpecificDetails and addition of Score).

src/main/riddl/Events/eventContext.riddl
JamesYopp: Ahh. I didn't understand that Duration was a defined type in RIDDL. In my mind a RIDDL spec is a business level specification. It doesn't mean that it is not precise. It means that the intended user thinks differently. Even though I spent a fair amount of time writing code and I have worked with engineering teams for 30 years I would not expect to provide a duration here in milliseconds. I don't know many people who would think of delaying an event in terms of milliseconds. They would think in terms of "I need to delay this event by a week", or "Because of rain, we are delaying the game by an hour".

JamesYopp (dismissed):

JamesYopp (approved):

PR #64, opened 2023-09-06
feat: Member -> User service [IA 249]

AlexWeinstein92:
Users instead of Members, where Member is a Role a User can take.

Currently speculative, needs to be reviewed by @JamesYopp and @reid-spencer and probably discussed, especially regarding what goes in `AdminInfo`, or if any restructuring is needed due to incorrect assumptions.

See https://yoppworks.atlassian.net/browse/IA-249 for more details


src/main/riddl/AllIdsProjections/allIdsProjections.riddl
```diff
@@ -9,8 +9,8 @@ result AllOrgIds is {orgs: OrganizationId*}
 query GetAllEventIds is {}
 result AllEventIds {events: EventId*}
 
-query GetAllMemberIds is {}
-result AllMemberIds is {members: MemberId*}
+query GetAllUserIds is {}
+result AllUserIds is {users: UserId*}
```
reid-spencer, line 13: As mentioned in our conversation, the only changes that are needed are these:
* Add a new User entity to represent all users of any role
* Add a "userId" field to Member to link it to the user that has that role
There should be no need to change these query or result names. The projection can just use the "userId" field from the various member instances

src/main/riddl/Applications/application.riddl
```diff
@@ -1,18 +1,20 @@
...
-    info: MemberContext.EditableInfo briefly "Details about an organization required for it to exist. See OrganizationContext.Organization.Info for details.",
+    info: OrganizationContext.EditableInfo briefly "Details about an organization required for it to exist. See OrganizationContext.Organization.Info for details.",
     tenant: TenantId,
     parent: Parent?,
+    admins: AdminList,
     members: MemberList,
     owners: OwnerList,
     contacts: ContactList,
-    onBehalfOf: MemberContext.MemberId briefly "The Id of the Member who is issuing the command."
+    onBehalfOf: UserContext.UserId briefly "The Id of the User who is issuing the command."
```
reid-spencer, line 10: I'm not sure this even has relevance as a necessary change. The Member has a userId but it also has its own unique Id.  It could be possible to just use the MemberId because we don't care about non-member users.

src/main/riddl/Applications/application.riddl
```diff
@@ -1,18 +1,20 @@
...
   } briefly "If parentOrg is not defined you are creating a BaseOrganization. If it is, you are creating a SubOrganization."
 
   result OrganizationEstablished is {
     organizationId: OrganizationId briefly "Unique identifier returned by EstablishOrganization command",
     info: OrganizationContext.Info,
     parent: Parent?,
+    admins: AdminList,
```
reid-spencer, line 17: I don't think this is necessary? Owner == Admin doesn't I?

src/main/riddl/Applications/application.riddl
```diff
@@ -44,35 +46,46 @@ application UserInterface is {
   }
 
   command RegisterMember is {
-    info: MemberContext.EditableInfo briefly "Details about a member required for it to exist. See MemberContext.Member.MemberInfo for details.",
-    onBehalfOf: MemberContext.MemberId
+    info: UserContext.EditableMemberInfo briefly "Details about a member required for it to exist. See UserContext.EditableMemberInfo for details.",
+    onBehalfOf: UserContext.UserId
   }
```
reid-spencer, line 51: No need to make this change.

src/main/riddl/Applications/application.riddl
```diff
@@ -44,35 +46,46 @@ application UserInterface is {
...
 
   result MemberRegistered is {
-    memberId: MemberContext.MemberId briefly "Unique identifier returned by RegisterMember command",
-    info: MemberContext.EditableInfo,
-    meta: MemberMetaInfo
+    memberId: UserContext.UserId briefly "Unique identifier returned by RegisterMember command",
+    info: UserContext.EditableMemberInfo,
+    meta: UserContext.UserMetaInfo
+  }
```
reid-spencer, line 57: No need to rename these.

reid-spencer (commented):
I stopped reviewing ½ way down because the Member ⇒ User renaming seems to be most of the work and is incorrect. See my first change.  We probably need to have another conversation about this patch.

